PR #5, opened 2016-10-13
Issue 3 - place overview carousel

fluffyemily:
Adding all the views required to render the place carousel

Quite a big commit, but I tried to break it down as much as possible for ease of review. We'll squash before merging.

Most of the added files are assets

#5


fluffyemily:
[Place Carousel](https://trello.com/c/MyRuC1k3/98-place-carousel)

Prox/Prox/PlaceCarousel/PlaceCarouselCollectionViewCell.swift
```diff
@@ -180,8 +193,40 @@ class PlaceCarouselCollectionViewCell: UICollectionViewCell {
     override func layoutSubviews() {
         super.layoutSubviews()
 
-        NSLayoutConstraint.activate([roundedBackgroundView.heightAnchor.constraint(equalToConstant: contentView.frame.size.height),
-                                     roundedBackgroundView.widthAnchor.constraint(equalToConstant: contentView.frame.size.width)])
+        var updatedConstraints = [roundedBackgroundView.heightAnchor.constraint(equalToConstant: contentView.frame.size.height),
+                                  roundedBackgroundView.widthAnchor.constraint(equalToConstant: contentView.frame.size.width)]
+
+        var deactivateConstraints = [NSLayoutConstraint]()
```
mcomella: Can you explain what's going on here? These more complex constraint combinations are confusing!

Prox/Prox/PlaceCarousel/PlaceCarouselCollectionViewCell.swift
```diff
@@ -0,0 +1,97 @@
...
+        return view
+    }()
+
+    lazy var placeImage: UIImageView = {
+        let imageView = UIImageView()
+        imageView.translatesAutoresizingMaskIntoConstraints = false
+        imageView.contentMode = .scaleAspectFill
+
+        let opacityView = UIView()
+        opacityView.backgroundColor = UIColor(colorLiteralRed: 0.0, green: 0.0, blue: 0.0, alpha: 0.1)
```
fluffyemily: What we are doing here is adding a slightly opaque dark masque to the top of the image, making it dark enough to see the white text on top, as opposed to making the image itself less opaque, which is what changing the alpha value of the `UIImageView` would do

Prox/Prox/PlaceCarousel/PlaceCarouselCollectionViewCell.swift
```diff
@@ -0,0 +1,97 @@
...
+                                        tripAdvisorReview.heightAnchor.constraint(equalToConstant: 50),
+                                        tripAdvisorReview.widthAnchor.constraint(equalToConstant: 100)])
+
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    override func layoutSubviews() {
+        super.layoutSubviews()
+
+        NSLayoutConstraint.activate([roundedBackgroundView.heightAnchor.constraint(equalToConstant: contentView.frame.size.height),
```
fluffyemily: yes, in setup views, the frame is set to zero making querying it for actual sizes impossible. We have to wait until the view are actually laid out before we get actual resolved view sizes. In `setupViews` we can sort out relative layout, but any constant layout that requires reference to the frame will have to be done in `layoutSubviews`.

Prox/Prox/Widgets/PlaceReviewView.swift
```diff
@@ -0,0 +1,80 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import UIKit
+import QuartzCore
+
+class PlaceReviewView: UIView {
```
fluffyemily: Naming has never been my strong point. Ask anyone, I'm almost famous for it.

Prox/Prox/Widgets/PlaceReviewView.swift
```diff
@@ -0,0 +1,80 @@
...
+    lazy var reviewScore: ReviewScoreView = {
+        let view = ReviewScoreView()
+        view.color = self.color
+        view.translatesAutoresizingMaskIntoConstraints = false
+        return view
+    }()
+
+    lazy var numberOfReviewersLabel: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
```
fluffyemily: Yeah, that is probably a good idea.

This method of using AutoLayout only came about in iOS9, but because on FxiOS we were supporting iOS8 we couldn't use it. Therefore I haven't yet come up with any 'tricks' to solve this problem. It is annoying as f*** to have to remember to add `translatesAutoresizingMaskIntoConstraints` to every view.

PR #7, opened 2016-10-13
Issue 6 - Create Podfile with inital required pods. Create workspace.…

fluffyemily:
… Update README with install/setup instructions

  #7


fluffyemily:
[integrate cocoapods (because Firebase does not support carthage)](https://trello.com/c/OpE7EUO2/114-integrate-cocoapods-because-firebase-does-not-support-carthage)

PR #12, opened 2016-10-14
Request location permission & add user location to map

mcomella:
We request location once when the app enters the foreground. In an ideal situation, this should be good enough for the app ("You're here! These are all the places nearby you!"). However, on a real device, this may not work:

* The initial location may be the first possible location the device can retrieve, which may not be the most accurate location the device can retrieve.
* The user's location may improve over-time so we may be able to provide better location data if we continuously monitor location.

I have yet to test this on a real device. I think this is good enough to start, however.

Eventually, we'll also want to monitor the user's location in background so we can provide place and event notifications.

This PR is built on an intermediate version of Emily's issue 3 branch and will need to be rebased before landing.

(Should I have an issue for this PR?)


mcomella:
[Choose Location API(s) that provide place type, proximity, and detailed info](https://trello.com/c/9dy2Ls7p/68-choose-location-api-s-that-provide-place-type-proximity-and-detailed-info)

mcomella:
[Detect & track users current location](https://trello.com/c/3oQUE677/96-detect-track-users-current-location)

Prox/Prox/AppDelegate.swift
```diff
@@ -38,6 +39,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func applicationWillEnterForeground(_ application: UIApplication) {
         // Called as part of the transition from the background to the active state; here you can undo many of the changes made on entering the background.
+        placeCarouselViewController!.refreshLocation()
```
fluffyemily: You will also need to add this line to `applicationDidBecomeActive`. When the app starts from cold, ` applicationDidBecomeActive` is called. When it is awoken suspension, then `applicationWillEnterForeground` is called. If you do not request user location in `applicationDidBecomeActive` then we will never get a request to monitor location from a new install.

Prox/Prox/LocationPermissions.swift
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+private func presentSettingsOrQuitPrompt(_ viewController: UIViewController) {
+    let appName = Bundle.main.object(forInfoDictionaryKey: "CFBundleName") as! String
+    let alertController = UIAlertController(title: "\(appName) requires location access",
+        message: "This prototype is not supported without location access.", preferredStyle: .alert)
+
+    let settingsAction = UIAlertAction(title: "Settings", style: .default, handler: {(action: UIAlertAction) -> Void in
+        UIApplication.shared.openAppSettings()
```
mcomella: That's a good point about conflicting with future iOS versions – could we get around it with namespacing, e.g. `UIApplication.shared.mozOpenAppSettings()`? I think there is value in making the code simpler to read and making programmer intent clearer.

That being said, I'll inline the calls for now.

Prox/Prox/AppDelegate.swift
mcomella: Removed from `applicationWillEnterForeground` and added to `DidBecomeActive`.

Prox/Prox/LocationPermissions.swift
```diff
@@ -0,0 +1,57 @@
...
+    })
+    let quitAction = UIAlertAction(title: "Quit", style: .destructive, handler: {(action: UIAlertAction) -> Void in
+        UIControl().suspend()
```
mcomella: Inlined (see above).

Prox/Prox/LocationPermissions.swift
```diff
@@ -0,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import MapKit
+import UIKit
+
+func hasLocationPermissionAndEnabled() -> Bool {
```
mcomella: Moving the globals to extension functions – good idea! Done.

mcomella:
I addressed the review comments in the topmost commit, exception to the `UIApplication`/`UIControl` where I merged the commit into the original commit because I thought it'd help with `pbxproj` rebasing but now I'm not so sure. Oh well.

Issue #8, opened 2016-10-13
Setup Firebase database


fluffyemily:
Integrate Firebase and setup a database that suits our querying strategy

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [Integrate Firebase https://www.firebase.com/docs/ios/quickstart.html](https://trello.com/c/yGaE9t9J/97-integrate-firebase-https-www-firebase-com-docs-ios-quickstart-html)

PR #13, opened 2016-10-14
Issue 8 firebase database setup

fluffyemily:
Add Firebase 3
Add Geofire (couldn't use cocoapods, Firebase SDK version and Geofire cocoapod were incompatible, so have included Geofire directly as thirdparty code)
Setup and authorise Firebase when app starts

#8 

Based of a #7


fluffyemily:
[Integrate Firebase](https://trello.com/c/yGaE9t9J/97-integrate-firebase)

mcomella:
> Firebase SDK version and Geofire cocoapod were incompatible, so have included Geofire directly as thirdparty code

I'm not sure I understand – so the Firebase SDK pod and the Geofire pod in Cocoapods are incompatible? That's concerning. I did a brief search and it seems people were having maybe related problems relatively recently ([link](https://github.com/firebase/geofire-objc/issues/48)).

Prox/Prox/AppDelegate.swift
```diff
@@ -23,9 +26,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
         return true
     }
 
+    private func setupFirebase() {
+        FIRApp.configure()
+
+        if let user = FIRAuth.auth()?.currentUser {
+            authorizedUser = user
+        } else {
+            FIRAuth.auth()?.signInAnonymously { (user, error) in
```
mcomella: Interesting – I only see [`signInAnonymouslyWithCompletion` in the docs](https://firebase.google.com/docs/reference/ios/firebaseauth/interface_f_i_r_auth.html#a0ffa43acd046e5ee42967106f73556f6) – am I looking at the right version of the docs?

mcomella:
Note the Buddybuild failure though!

Prox/Prox/AppDelegate.swift
fluffyemily: yes you are. It's an interesting feature of swift, if the final argument of a function is a closure, you can define that closure at the end of the function, not inside it, and you don't have to specify it's var name. This _is_  [`signInAnonymouslyWithCompletion`](https://firebase.google.com/docs/reference/ios/firebaseauth/interface_f_i_r_auth.html#a0ffa43acd046e5ee42967106f73556f6)

Issue #17, opened 2016-10-18
Fetch Travel Times from MapKit


fluffyemily:
Use MapKit's `MKDirections` object to fetch travel times from the users current location to a given set of Coordinates

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [Fetch Travel Times for Places from Current Location](https://trello.com/c/Vu0uy7xn/143-fetch-travel-times-for-places-from-current-location)

PR #18, opened 2016-10-18
Issue 17 - travel times

fluffyemily:
#17 

Use MapKit's inbuilt routing ability to route from one set of coordinates to another using a specified transport type (walking, driving or public transport)


fluffyemily:
[Fetch Travel Times for Places from Current Location](https://trello.com/c/Vu0uy7xn/143-fetch-travel-times-for-places-from-current-location)

Prox/Prox/Data/TravelTimesProvider.swift
```diff
@@ -0,0 +1,54 @@
...
+struct TravelTimesProvider {
+
+    static func travelTime(fromLocation: CLLocationCoordinate2D, toLocation: CLLocationCoordinate2D, byTransitType transitType: MKDirectionsTransportType = .any, withCompletion completion: @escaping ((TravelTimes?) -> ())) {
+
+        let directionsRequest = MKDirectionsRequest()
+        if #available(iOS 10.0, *) {
+            directionsRequest.source = MKMapItem(placemark: MKPlacemark(coordinate: fromLocation))
+            directionsRequest.destination = MKMapItem(placemark: MKPlacemark(coordinate: toLocation))
+        } else {
+            directionsRequest.source = MKMapItem(placemark: MKPlacemark(coordinate: fromLocation, addressDictionary: nil))
```
mcomella: nit: It seems simpler if we used this initializer for all versions – is there an advantage to using the iOS 10+ initializer?

Prox/ProxTests/TravelTimesTests.swift
```diff
@@ -0,0 +1,100 @@
...
+        super.tearDown()
+    }
+    
+    func testAnyTravelTimes() {
+
+        let waitQuery = expectation(description: "Waiting for travel times to be calculated")
+
+        TravelTimesProvider.travelTime(fromLocation: testSource, toLocation: testDestination) { travelTimes in
+            XCTAssertNotNil(travelTimes)
+            XCTAssertNil(travelTimes?.walkingTime)
```
mcomella: We didn't provide a specific travel type – how do we know for sure this will be nil?

Prox/ProxTests/TravelTimesTests.swift
```diff
@@ -0,0 +1,100 @@
...
+            XCTAssertNotNil(travelTimes?.drivingTime)
+            XCTAssertNil(travelTimes?.publicTransportTime)
+            waitQuery.fulfill()
+        }
+
+        waitForExpectations(timeout: 5.0)
```
mcomella: We're making a network request here, right? This seems like a short timeout (though if we pay for buddybuild by cpu time maybe it's a good one) – I'd think 30 seconds would probably be reasonable.

Prox/ProxTests/TravelTimesTests.swift
```diff
@@ -0,0 +1,100 @@
...
+            XCTAssertNotNil(travelTimes?.publicTransportTime)
+            waitQuery.fulfill()
+        }
+
+        waitForExpectations(timeout: 5.0)
+        
+    }
+
+    func testInvalidRoute() {
+        let testInvalidDestination = CLLocationCoordinate2D(latitude: 19.9542305, longitude: 155.8531072)
```
mcomella: Let's hope they never build a bridge... 😜

Prox/Prox/Data/TravelTimesProvider.swift
```diff
@@ -45,10 +49,33 @@ struct TravelTimesProvider {
             completion(travelTime)
         }
     }
+
+    static func travelTime(fromLocation: CLLocationCoordinate2D, toLocation: CLLocationCoordinate2D, byTransitTypes transitTypes: [MKDirectionsTransportType], withCompletion completion: @escaping ((TravelTimes?) -> ())) {
+        var allTimes = [TravelTimes?]()
+        for transitType in transitTypes {
```
mcomella: fwiw, I added the Deferred lib in [my branch](https://github.com/mcomella/prox/tree/server-comm-deferred/) but given this is a prototype, it's probably not worth changing something that works.

Note that the Deferred lib crashed with `.allFilled` (equivalent to Swift 2's `.all`) when returning optionals so it may not be usable here.

mcomella: Having all the travel time data keeps things simple when we need to choose what to display the user. 👍 

However, if we make these requests in quick succession, as we do here, would they rate limit us?

Prox/Prox/Data/TravelTimesProvider.swift
fluffyemily: the only advantage is if they decide to deprecate the older one at a future date. As that is a future date rather than immediate, I'll make the change you suggest

Prox/ProxTests/TravelTimesTests.swift
fluffyemily: I guess we don't know _for sure_ but the order that `MapKit` prioritises transport types is `automobile`, `transit`, `walking`, so if it is possible to route from `testSource` to `testDestination` by car then that will be the one that is filled. I'll turn this into a check that at least one of the values is filled.

Prox/ProxTests/TravelTimesTests.swift
fluffyemily: I would rather leave the wait time short until we figure out that we need longer. We don't really want our users waiting much longer than 5 seconds for their routes, and so in a way this is a little performance test too....

PR #21, opened 2016-10-19
Query the database and insert results into the UI

mcomella:
There are many TODOs and it's not very robust but this may allow us to better break up the remaining work to do.

@fluffyemily, if you don't feel we should merge this yet, let me know.

### DIFF
--- a/Prox/Prox/Models/Place.swift
+++ b/Prox/Prox/Models/Place.swift
@@ -3,6 +3,10 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
+import Firebase
+
+private let PROVIDERS_PATH = "providers/"
+private let YELP_PATH = PROVIDERS_PATH + "yelp"
 
 class Place {
 
@@ -25,6 +29,7 @@ class Place {
 
     // todo: hours
 
+    // TODO: temporary? for testing purposes.
     init(name: String,
          categories: [String],
          url: String,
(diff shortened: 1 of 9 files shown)


Prox/Prox/Models/Place.swift
```diff
@@ -3,7 +3,87 @@
...
+        self.photoURLs = photoURLs
+    }
+
+    init?(fromFirebaseSnapshot data: FIRDataSnapshot) {
+        guard data.exists(), data.hasChildren(), let value = data.value as? NSDictionary else {
+            return nil
+        }
+
+        // TODO: handle missing values more robustly
+        self.name = value["id"] as? String ?? "Name unknown"
```
mcomella: Good call! I'm used to not having optionals so I tend towards immutable, default values – let me know if you see other places where optionals can come in handy!

Prox/Prox/Models/Place.swift
```diff
@@ -3,7 +3,87 @@
...
+        self.summary = value["pullQuote"] as? String ?? "Summary unknown"
```
mcomella: I don't know what the difference is between `description` and `pullQuote` so I'm just going to leave it as `pullQuote` for now.

Prox/Prox/Models/Place.swift
```diff
@@ -3,7 +3,87 @@
...
+
+        self.address = (value["address"] as? [String])?.joined(separator: " ") ?? "Address unknown"
+        if let coords = value["coordinates"] as? [String:Double],
+            let lat = coords["lat"], let lon = coords["lon"] {
+            self.latLong = CLLocationCoordinate2D(latitude: lat, longitude: lon)
+        } else {
+            // TODO: if we already have the address, maybe we reverse-geocode & do better. In any
```
mcomella: Not at all! I'd rather be corrected now than to be embarrassed later!

...I figured I had a 50/50 shot of getting it right, anyway. :P

Prox/Prox/Models/ReviewProvider.swift
```diff
@@ -0,0 +1,44 @@
...
+import Foundation
+import Firebase
+
+class ReviewProvider {
+
+    let url: String
+
+    let rating: Double
+    let reviews: [String]
+    let totalReviewCount: Int
```
mcomella: To be explicit, on `ReviewProvider`, I made `url` required and all other fields (`rating`, `reviews`, `totalReviewCount`) as optional.

mcomella: I made `url` required because I figured we really need to be able to attribute the reviews and, if we can't, it's better that we say, "Data not available" than show data that's linkable.

That being said, users may prefer to see the data even if we can't link directly, in which case, I suppose we could just change it.

mcomella:

mcomella:
[Determine front-end data structures](https://trello.com/c/7NEPcbXy/130-determine-front-end-data-structures)

mcomella:

Issue #15, opened 2016-10-17
Review Scores can be floats


fluffyemily:
Currently the review scores only allow for whole integer values, whereas they should be able to handle float values. Update the ReviewScoreView to render and handle float values.

* Change Int to Float in instance variables
* Update rendering code to allow for partial fills for score shapes

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [Frontend Bug Fixing](https://trello.com/c/Ac31lPnL/144-frontend-bug-fixing)

PR #26, opened 2016-10-20
Issue 15 - redraw review score view as 2 halves so we can render floa…

fluffyemily:
…t values.

Update scores to be floats and update the test values in PlaceCarousel to test half score rendering

#15


Prox/Prox/Widgets/ReviewScoreView.swift
```diff
@@ -41,9 +36,9 @@ class ReviewScoreView: UIStackView {
 
         for index in 0..<5 {
             let scoreItem = ReviewScoreItemView()
-            scoreItem.color = color
             scoreItem.backgroundColor = .clear
-            scoreItem.fill = index < score
+            let scoreLeft = self.score - Float(index)
+            setFillScore(score: scoreLeft, forView: scoreItem)
```
mcomella: Rather than duplicating this code, should we call `updateReviewScore` after the for loop?

Prox/Prox/Widgets/ReviewScoreView.swift
```diff
@@ -62,34 +57,58 @@ class ReviewScoreView: UIStackView {
...
+            view.rightColor = Colors.reviewScoreDefault
+        }
+    }
+
 }
 
 private class ReviewScoreItemView: UIView {
 
-    var color: UIColor? {
+    var leftColor: UIColor? {
```
mcomella: It might be slightly safer to have:

```
private enum FillAmount {
  case empty, half, filled
}

private class ReviewScoreItemView: UIView {
  let color: UIColor // defined in init
  var fillAmount: FillAmount = .empty // set from `updateReviewScore`/`setFillScore`
...
```
But if it already works, it's probably not worth the time.

Prox/Prox/Widgets/ReviewScoreView.swift
```diff
@@ -62,34 +57,58 @@ class ReviewScoreView: UIStackView {
         }
     }
 
+    private func updateReviewScore(score: Float) {
+        for (index, subview) in arrangedSubviews.enumerated() {
+            if let scoreItem = subview as? ReviewScoreItemView {
+                let scoreLeft = self.score - Float(index)
```
mcomella: nit: -> `scoreRemaining` or `scoreLeftOver`

I kept thinking left as in left/right.

Prox/Prox/Widgets/ReviewScoreView.swift
```diff
@@ -62,34 +57,58 @@ class ReviewScoreView: UIStackView {
...
+                setFillScore(score: scoreLeft, forView: scoreItem)
+            }
+        }
+    }
+
+    private func setFillScore(score: Float, forView view: ReviewScoreItemView) {
```
mcomella: This could be a method on `ReviewScoreItemView` which (as per my comment below) might be better suited to take a `FillAmount` enum rather than a Float, but if it works now, it's probably fine.

Prox/Prox/Widgets/ReviewScoreView.swift
```diff
@@ -62,34 +57,58 @@ class ReviewScoreView: UIStackView {
...
-            } else {
-                context.setFillColor(Colors.reviewScoreDefault.cgColor)
-            }
-            context.fillEllipse(in: rect)
-        }
+        let width = rect.size.width / 2
+
+        let leftHalf = UIBezierPath(roundedRect: CGRect(x: rect.origin.x, y: rect.origin.x, width: width, height: rect.size.height), byRoundingCorners: [UIRectCorner.topLeft, UIRectCorner.bottomLeft], cornerRadii: CGSize(width: 10, height: 10))
+        leftHalf.close()
+        (leftColor ?? UIColor.black).setFill()
```
mcomella: Why `UIColor.black`? We previously set `Colors.reviewScoreDefault` – should we make that the default? Is it for uninitialized errors? In which case I'd recommend throwing.

Prox/Prox/Widgets/ReviewScoreView.swift
fluffyemily: I thought about that, but then we would be looping through the views twice, whereas here we only loop through them once. For efficiency I prefer it this way.

Prox/Prox/Widgets/ReviewScoreView.swift
mcomella: > It was just so that we had a default color in case no one set a color

I prefer to crash when the app is in an unexpected state – assuming you can get crash reports, you can find errors early and fix them, rather than having an error which may be a slightly less bad experience but that you may never fix. This is particularly important when errors may compound (e.g. one view being black is alright but when all the views on the screen are black).

Thinking about it, it's less important for UI but more so for data processing – which is perhaps similar to your point about ViewControllers throwing the fit rather than the views.

Issue #29, opened 2016-10-20
App Crashing when unable to create a place from current data


fluffyemily:
When Place creation returns a nil, the app is killed deliberately causing all hell to break loose. Make this case handle more gracefully somehow

PR #30, opened 2016-10-20
Issue 29 - crash on empty place

fluffyemily:
#29 

Problem: App was force exiting when unable to create a place successfully.
Solution: Return a PlacesResult object that encapsulates success or failure that the receiver can use to make informed decisions

Possible Improvements:

Make the Result object generic so it can handle more than just Places (could be handy when we start also fetching events).


Prox/Prox/Data/PlacesResult.swift
```diff
@@ -0,0 +1,51 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+private enum DeferredResult {
+    case success
+    case failure
+}
+
+class PlacesResult {
```
mcomella: nit: I'd add a comment to explain why we do this. Maybe take the one from before?

Prox/Prox/PlaceCarouselViewController.swift
```diff
@@ -144,7 +144,13 @@ extension PlaceCarouselViewController: CLLocationManagerDelegate {
             // Make sure we only call this once, for testing purposes.
             if !once {
                 FirebasePlacesDatabase().getPlaces(forLocation: TEST_LL).upon(DispatchQueue.main) { places in
-                    self.placeCarousel.places = places.flatMap { $0 as? Venue }
+                    self.placeCarousel.places = places.flatMap { result in
+                        if result.isSuccess() {
```
mcomella: We could remove the complexity here if `PlacesResult.result` just returned an optional and we let flatMap take care of removing the bad entries.

But it works and it's a prototype so it's probably not worth changing now.

Prox/Prox/Data/PlacesResult.swift
```diff
@@ -9,43 +9,43 @@ private enum DeferredResult {
     case failure
 }
 
-class PlacesResult {
+class DatabaseResult<T> {
```
mcomella: nit: -> `DeferredResult` or `DeferredResultWrapper`? I think we can use this for more than just Database results.

Prox/Prox/Data/DatabaseResult.swift
```diff
@@ -4,6 +4,16 @@
 
 import Foundation
 
+// This is an empty class that forms a base class for all places
```
mcomella: Might want to update this to be a little clear (or if we don't want to spend the time, add a brief comment saying this is not entirely accurate but you get the idea ;)).

PR #31, opened 2016-10-20
Issue 28 integrate travel times

fluffyemily:
Display the travel times from users current location to a place on the place carousel


Prox/Prox/PlaceCarousel/PlaceCarousel.swift
```diff
@@ -70,17 +72,49 @@ extension PlaceCarousel: UICollectionViewDataSource {
...
     }
 
+    private func setTravelTimes(travelTimes: TravelTimes?, forCell cell: PlaceCarouselCollectionViewCell) {
+        guard let travelTimes = travelTimes else {
+            return
+        }
+
+        if let walkingTimeSeconds = travelTimes.walkingTime {
+            let walkingTimeMinutes = Int(round(walkingTimeSeconds / 60.0))
+            if walkingTimeMinutes <= 30 {
```
mcomella: nit: 30 as a constant – I assume we may want to configure this (perhaps dynamically based on location).

Prox/Prox/PlaceCarousel/PlaceCarousel.swift
```diff
@@ -70,17 +72,49 @@ extension PlaceCarousel: UICollectionViewDataSource {
...
+                if walkingTimeMinutes < 3 {
+                    cell.locationImage.image = UIImage(named: "icon_location")?.withRenderingMode(UIImageRenderingMode.alwaysTemplate)
+                    cell.location.text = "You're here"
```
mcomella: We may want to make sure this only appears on 1 item (or maybe multiple if they're combined, e.g. hotel bar & pool as two different places). We can figure that out when we get the real data, however.

Prox/Prox/PlaceCarousel/PlaceCarousel.swift
```diff
@@ -70,17 +72,49 @@ extension PlaceCarousel: UICollectionViewDataSource {
...
+                    cell.isSelected = true
```
mcomella: Since we may get more than one location that we're at, I'm not sure we want to use that as the selection criteria.

Prox/Prox/PlaceCarousel/PlaceCarousel.swift
```diff
@@ -70,17 +72,49 @@ extension PlaceCarousel: UICollectionViewDataSource {
...
+                return
+            }
+        }
+
+        if let drivingTimeSeconds = travelTimes.drivingTime {
+            let drivingTimeMinutes = Int(round(drivingTimeSeconds / 60.0))
+            cell.location.text = "\(drivingTimeMinutes) min drive away"
+            cell.isSelected = false
+        } else {
+            return
```
mcomella: nit: `else { return }` is unnecessary.

Prox/Prox/PlaceCarousel/PlaceCarousel.swift
fluffyemily: Yeah, we can define this in the Remote Config once we've set that up, but then we can move it into user configuration when we start putting settings in post Hawaii.

Prox/Prox/PlaceCarousel/PlaceCarousel.swift
fluffyemily: :+1: Simplest possible solution first, then fine tune as we know what we need

Prox/Prox/PlaceCarousel/PlaceCarousel.swift
fluffyemily: again, let's figure that our as we start to understand our data more

PR #33, opened 2016-10-20
Clean up FirebasePlacesDatabase code & TODOs.

mcomella:


mcomella:

Prox/Prox/FirebasePlacesDatabase.swift
```diff
@@ -26,17 +26,14 @@ class FirebasePlacesDatabase: PlacesDatabase {
...
-    // todo: handle errors (double-check every query)
-    // TODO: handle version in DB
     /*
      * Queries GeoFire to get the place keys around the given location and then queries Firebase to
      * get the place details for the place keys.
      */
     func getPlaces(forLocation location: CLLocation) -> Future<[DatabaseResult<Place>]> {
-        // TODO: OperationQueue.current okay?
-        let places = getPlaceKeys(aroundPoint: location).andThen(upon: OperationQueue.current!) { (placeKeyToLoc) -> Future<[DatabaseResult<Place>]> in
+        let queue = DispatchQueue.global(qos: .userInitiated)
```
fluffyemily: it's really interesting seeing you coming straight to swift 3.0 / iOS10 without being encumbered with the knowledge of how things used to work in swift 2.x / previous iOS versions. I had no idea about the `qos` option on `DispatchQueue` :)

Prox/Prox/FirebasePlacesDatabase.swift
```diff
@@ -88,12 +84,15 @@ class FirebasePlacesDatabase: PlacesDatabase {
 
         let childRef = placeDetailsRef.child(placeKey)
         childRef.queryOrderedByKey().observeSingleEvent(of: .value) { (data: FIRDataSnapshot) in
+            guard data.exists() else {
```
fluffyemily: We could probably make this a little more concise like this
```
            guard data.exists(),
                let place = Place(fromFirebaseSnapshot: data) else {
                    let message = data.exists() ? "Child place key does not exist: \(placeKey)" : "Snapshot missing required Place data: \(data)"
                    deferred.fill(with: DatabaseResult.fail(withMessage: message))
                    return

            }
            deferred.fill(with: DatabaseResult.succeed(value: place))
```
But seeing as it works just fine this way, I see no reason to change it.

fluffyemily:
@mcomella I am merging this so that it's all set for you when you get in

PR #39, opened 2016-10-21
Update place count in UI

mcomella:


mcomella:

Prox/Prox/PlaceCarouselViewController.swift
```diff
@@ -218,7 +226,7 @@ extension PlaceCarouselViewController: CLLocationManagerDelegate {
         mapView.region = MKCoordinateRegion(center: center, span: span)
 
         FirebasePlacesDatabase().getPlaces(forLocation: location).upon(DispatchQueue.main) { places in
-            self.placeCarousel.places = places.flatMap { $0.successResult() }
+            self.places = places.flatMap { $0.successResult() }
```
fluffyemily: we could set the text for the `numberOfPlacesLabel` here and then pass the places on to the `PlaceCarousel` for storage.

```
self.placeCarousel.places = places.flatMap { $0.successResult() }
headerView.numberOfPlacesLabel.text = "\(places.count) places"
```

mcomella: @fluffyemily What are your concerns with my implementation? Is it duplicating the storage?

I'm most interested in updating the UI from a single place. Ideally, this is `places.didSet` – if the UI is updated directly from the data structure, you avoid issues where the UI doesn't match the backing data (caveat: I don't know how well this style works with animations).

Two solutions that take both sentiments into account:
1. Pass a closure getter to `places` in `placeCarousel.init`. This lets placeCarousel access places directly without duplicating the data.
2. Move the dataSource delegate to `PlaceCarouselViewController`. Since the ViewController should be controlling the views, which arguably means providing their data, this encapsulation makes sense, at the cost of a larger file. 

I chose not to use these solutions because of their added complexity and thus difficulty to later change but we could use them if you prefer. I prefer 1).

fluffyemily: I agree that your first option is better, but instead of using a closure, we should create a delegate protocol to provide the data to the `PlaceCarousel`. That same delegate can then be used to provide data to the place details screens that we are using in https://github.com/mozilla-mobile/prox/tree/fluffyemily/issue-43-open-detail-view-from-place-carousel

mcomella:
Update in favor of #55.

PR #40, opened 2016-10-21
Download and insert place carousel images.

mcomella:
The image downloader will need to be moved in order for the place details card
to access it (to use the same cache).

That being said this is a first run to get something in so that we have
something to move forward with when designing the place details.

(May merge conflict with my places label PR – github makes it hard to build a branch on top of another one and keep the reviews clean 😢)

### DIFF
--- a/Prox/Prox/PlaceCarousel/PlaceCarousel.swift
+++ b/Prox/Prox/PlaceCarousel/PlaceCarousel.swift
@@ -4,7 +4,7 @@
 
 import Foundation
 import UIKit
-
+import AFNetworking
 
 private let CellReuseIdentifier = "PlaceCarouselCell"
 private let MIN_WALKING_TIME = 30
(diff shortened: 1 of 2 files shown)


mcomella:

Prox/Prox/PlaceCarousel/PlaceCarousel.swift
```diff
@@ -85,9 +99,34 @@ extension PlaceCarousel: UICollectionViewDataSource {
...
+                print("lol urlRes unexpectedly null")
+                return
+            }
+
+            guard res.statusCode == 200 else {
+                print("lol image status code unexpectedly \(res.statusCode)")
+                return
+            }
+
+            cell.placeImage.image = img
```
fluffyemily: are you sure we're on the main thread here?

fluffyemily: Ignore that, yes we are. `AFImageDownloader` takes care of that :)

mcomella:
> Not the way I would have done it - I would probably have made a subclass of UIImageView and created a downloadImageFromURL function that handled all of the AFNetworking, downloading and setting of the image,

That seems like a good way to encapsulate the functionality - good idea! I'll leave it for now but when we need to update it, I'll consider that approach.

Issue #41, opened 2016-10-24
Fake Location to Hawaii for buddybuild builds


fluffyemily:
When building apps for distrubution through buddybuild, fake the location so that the app can be properly tested outside of Hawaii

mcomella:
@fluffyemily Closed with #42, right?

fluffyemily:
Yup

PR #42, opened 2016-10-24
Build with location spoofing on device

fluffyemily:
#41 

Currently, builds installed via buddybuild show no data as we no longer spoof our location. Here, I've added the ability to do feature flags based on the current build configuration and then used that to create a variable that we use to determine whether or not to fake our location.

In future this could be extended to return a random Hawaii location, rather than just the Hilton Waikaloa so that we can check that the app works in various places around Kona


Prox/Prox/Utilities/AppConstants.swift
```diff
@@ -0,0 +1,55 @@
...
+        #elseif MOZ_CHANNEL_RELEASE
+            return AppBuildChannel.Release
+        #else
+            return AppBuildChannel.Developer
+        #endif
+    }()
+
+    /// Flag indiciating if we are running in Debug mode or not.
+    public static let isDebug: Bool = {
+        #if MOZ_CHANNEL_DEBUG
```
mcomella: nit: I feel like we should re-use `BuildChannel` here and below, e.g. `return BuildChannel == .Developer`:
* It re-uses existing code that we (hopefully) know works, which means it can be slightly less error-prone.
* If the configuration changes (e.g. we remove the developer channel), we're more likely to catch that error in `isDebug`.

Prox/Prox/PlaceCarouselViewController.swift
```diff
@@ -198,13 +198,18 @@ extension PlaceCarouselViewController: CLLocationManagerDelegate {
...
+        if var location = locations.last {
             // In iOS9, didUpdateLocations can be unexpectedly called multiple
             // times for a single `requestLocation`: we guard against that here.
             let now = Date()
             if timeOfLastLocationUpdate == nil ||
                 (now - MIN_SECS_BETWEEN_LOCATION_UPDATES) > timeOfLastLocationUpdate! {
                 timeOfLastLocationUpdate = now
+
+                if AppConstants.MOZ_LOCATION_FAKING {
+                    location = CLLocation(latitude: 19.9263136, longitude: -155.8868328)
```
mcomella: nit: I'd add a comment (or use a constant) to explain what this location is.

Prox/Prox/Utilities/AppConstants.swift
```diff
@@ -52,4 +52,17 @@ public struct AppConstants {
             return false
         #endif
     }()
+
+    // Enables/disables location faking for Hawaii
+    public static let MOZ_LOCATION_FAKING: Bool = {
+        #if MOZ_CHANNEL_DEBUG
```
mcomella: nit: As before, I feel like we should re-use the functions in this file.

Podfile.lock
```diff
@@ -73,4 +73,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 6ee5bcd18f0d60a4b82be64479784270153e3f26
 
-COCOAPODS: 1.1.0
+COCOAPODS: 1.0.1
```
mcomella: How do you want to resolve this? Should you update to 1.1.0 or should I see if I can downgrade? Maybe there's a way to create a virtualenv?

fluffyemily: I'll update. I thought I had, but maybe it didn't work

fluffyemily: OK, so updated cocoapods and am now on 1.1.1, so everything is different again! Jeez.

PR #52, opened 2016-10-25
Issue 44 - open closest place by default

fluffyemily:
For the purposes here I have decided to open the closest place according to as-the-crow-flies distance, but we may want to think about sorting by travel time.

* Sort places according to distance when received from `PlacesDatabase`
* When places array is set, if there were previously no places that we were showing, check to see whether we are currently displaying place details and open closest place if we are not.


fluffyemily:

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -41,6 +42,10 @@ class PlaceCarouselViewController: UIViewController {
         didSet {
             // TODO: how do we make sure the user wasn't interacting?
             placeCarousel.refresh()
+
+            if oldValue.count == 0 {
```
mcomella: I'm a little concerned that we might hit an edge case with this but it's good enough for now. :)

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -249,7 +273,7 @@ extension PlaceCarouselViewController: CLLocationManagerDelegate {
         mapView.region = MKCoordinateRegion(center: center, span: span)
 
         FirebasePlacesDatabase().getPlaces(forLocation: location).upon(DispatchQueue.main) { places in
-            self.places = places.flatMap { $0.successResult() }
+            self.places = PlaceUtilities.sort(places: places.flatMap { $0.successResult() }, byDistanceFromLocation: location)
```
mcomella: I'm not sure we want to sort *all* places by absolute distance but works for now.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -0,0 +1,20 @@
...
+    static func sort(places: [Place], byDistanceFromLocation location: CLLocation, ascending: Bool = true) -> [Place] {
+        return places.sorted { (placeA, placeB) -> Bool in
+            let placeADistance = location.distance(from: CLLocation(latitude: placeA.latLong.latitude, longitude: placeA.latLong.longitude))
+            let placeBDistance = location.distance(from: CLLocation(latitude: placeB.latLong.latitude, longitude: placeB.latLong.longitude))
+
+            if ascending {
+                return placeADistance < placeBDistance
+            }
+
+            return placeBDistance < placeADistance
```
mcomella: nit: I'd prefer the variables to keep the same order and switch the operator (`placeADistance > placeBDistance`). It's a little easier to scan.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: Yeah, I kinda figured that we would want to fine tune this when we start playing with the app and working out exactly what the logic should be

Issue #38, opened 2016-10-21
Design Error States for Home View


vazquez:
What if we can't retrieve:
- Description (from Wikipedia or Yelp)
- Ratings and Reviews
- Images
- URL
- Categories
- Closing Time

We will always be able to retrieve Venue Name and Transportation

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [I want to receive an actionable overview of relevant information about places near me, so that my day is not ruined with FOMO](https://trello.com/c/VPWm03rY/49-i-want-to-receive-an-actionable-overview-of-relevant-information-about-places-near-me-so-that-my-day-is-not-ruined-with-fomo)

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user I want to see a useful carousel card even if all the data is not present so that I can continue to the most out of my experience](https://trello.com/c/EmrIgDgb/166-as-a-user-i-want-to-see-a-useful-carousel-card-even-if-all-the-data-is-not-present-so-that-i-can-continue-to-the-most-out-of-my-)

vazquez:
![edge case ui](https://cloud.githubusercontent.com/assets/1119821/19970918/e4cf9442-a1b3-11e6-8c9b-c2f8d27cb4a6.png)

mcomella:
Thoughts:
* What about places without hours? e.g. a beach (which is probably closed at sunset, but yelp doesn't list hours)
* If we don't have a yelp rating, it'd probably be pretty hard to sort the places.
  * I spoke with Emily: we'll only figure out if this is a problem once we do the prioritization ([see trello card](https://trello.com/c/IBEHIXNb/149-as-a-user-i-want-to-see-the-most-useful-places-when-i-open-the-app-so-that-i-can-make-better-decisions-about-my-next-step)), so we won't implement this error state until we finish that card
* We actually won't have the website for a lot of these places :\ (unless we find & insert them by hand, which is a viable option).
  * We need factual places for the websites and not every yelp place has a factual place

mcomella:
We don't handle:
* No yelp reviews available
* No yelp ratings available

But let's postpone those because this data is important to prioritize places (see the previous comment).

I'll get to work on filtering out:
* Places without yelp ratings *and* reviews
* Places without images.

mcomella:
I don't show the place if we're missing either (or both) a yelp rating or a review. This isn't to the mocks, but see above.

[...]

PR #55, opened 2016-10-25
Match latest server schema & related goo

mcomella:


mcomella:

Prox/Prox/PlaceCarousel/PlaceCarousel.swift
```diff
@@ -87,7 +87,7 @@ extension PlaceCarousel: UICollectionViewDataSource {
         if let yelp = place.yelpProvider {
             // TODO: handle optionals in UI
             cell.yelpReview.score = Float(yelp.rating ?? -1)  // TODO: type
-            cell.yelpReview.numberOfReviewersLabel.text = "\(yelp.totalReviewCount!) Reviews"
+            cell.yelpReview.numberOfReviewersLabel.text = "\(yelp.totalReviewCount ?? -1) Reviews"
```
fluffyemily: Not sure I like the idea that we might tell the user there are -1 reviews. I'd rather default to 0 if we default to a number.

Alternatively we could set the text to nil in which case no review count will be displayed. Maybe @vazquez  has a preference considering his work from https://github.com/mozilla-mobile/prox/issues/38

PR #64, opened 2016-10-28
Create incomplete place details card

mcomella:
Landing this will:
* Reduce the burden of the next review
* Provide us the opportunity to tag team the UI (if there are enough cycles)


mcomella:

fluffyemily:
Have you thought about using a `UIStackView` for the `PlaceDetailsCardView`? It might make some of your layout much easier. If you make it a vertical stack view, you can just add the views (you would have to wrap walking/opening times views, and the review views in a single view) to the stack and let it sort out the rest of the layout. I think it could help a lot with making it a bit more simple.

Also, the `UIStackView` will adapt to changes in view height, so for the opening and closing of the summary and review views, all you would need to do is show/hide the appropriate views to have the whole thing adapt.

The only reason I didn't use that for the carousel view was the shadows. Views inside a `UIStackView` cannot then apply a shadow outside their bounds. But from what I can see with the `PlaceDetailsCardView` there are no shadows inside that view that would be affected by this, so makes it an ideal choice.

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -0,0 +1,164 @@
...
+        let view = UILabel()
+        view.textColor = Colors.detailsViewCardPrimaryText
+        view.font = Fonts.detailsViewCategoryText
+        return view
+    }()
+
+    // TODO: Set styling.
+    lazy var urlLabel: UILabel = {
+        let view = UILabel()
+        view.textColor = .blue
```
fluffyemily: I know this is a default color, but I would still prefer to add it as a constant inside `Colors` so that it becomes super easy to change if necessary.

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -0,0 +1,164 @@
...
+    }()
+
+    lazy var hoursView: PlaceDetailsIconInfoView = {
+        let view = PlaceDetailsIconInfoView()
+        view.iconView.image = UIImage(named: "icon_times")
+        return view
+    }()
+
+    lazy var yelpReviewView: ReviewContainerView = {
+        let view = ReviewContainerView(color: Colors.yelp, mode: .detailsView)
```
fluffyemily: If things like font, color and padding are changing between details and carousel views, I would rather those values became variables on the `ReviewContainerView` and were set explicitly inside the `UIViewController`. If a view is to be used in more than one way, I would feel happier if it didn't encapsulate that knowledge inside itself, but each `UIViewController` encapsulated the knowledge about how it needs it's views configured.

Prox/Prox/Widgets/ReviewContainerView.swift
```diff
@@ -5,8 +5,16 @@
 import UIKit
 import QuartzCore
 
+enum ReviewContainerViewMode {
```
fluffyemily: see comment earlier about encapsulating specific configurations inside the `UIViewController` rather than inside the `ReviewContainerView`. This enum should be removed and the constants below turned into variables that can be configured externally.

mcomella:
> Have you thought about using a UIStackView for the PlaceDetailsCardView?

I have not – thanks for the feedback! If we need to refactor the PlaceDetailsCardView, I'll consider it!

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
mcomella: Good point. I prefer things to be explicit and immutable, which is ideally done through the constructor. I used `mode` as a way of avoiding passing in a large number of values but I see now that perhaps I should have passed in a struct!

Issue #65, opened 2016-10-28
nchapman@mozilla.com on build #130: This text is low res.


sarentz:
Feedback from nchapman@mozilla.com : This text is low res.

<img src="https://s3-us-west-2.amazonaws.com/buddybuild-screenshots/57f51f8bcaa70a01005412ee/581344efe76efc0100b136aa/e48a29be-9b36-4dcd-a5c2-7ee1959c5652.jpg" width="33%" height="33%" /><table><tr><td>Created</td><td>Fri Oct 28 2016 13:18:30 GMT+0000 (UTC)</td></tr><tr><td>CFBundleShortVersion</td><td>1.0</td></tr><tr><td>CFBundleVersion</td><td>1</td></tr><tr><td>App uptime</td><td>132.1893</td></tr><tr><td>Build</td><td>130</td></tr><tr><td>Device type</td><td>iPhone 7 Plus</td></tr><tr><td>Device name</td><td>Nick's iPhone</td></tr><tr><td>Screen size</td><td>414</td></tr><tr><td>Screen size</td><td>414px by 736px</td></tr><tr><td>Battery</td><td>100% Unplugged</td></tr><tr><td>Memory free</td><td>178 MB / 2735 MB</td></tr><tr><td>Disk free</td><td>79305 MB / 116121 MB</td></tr><tr><td>Network IP</td><td>192.168.7.205</td></tr></table>

 [Link to buddybuild feedback from build 130](https://dashboard.buddybuild.com/apps/57f51f8bcaa70a01005412ee/feedback?fid=5813502676b29a01003824b5&bnum=130)

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [This text is low res.](https://trello.com/c/9zBaCy86/177-this-text-is-low-res)

PR #75, opened 2016-11-01
Issue 65 - Ensure text is not blurred

fluffyemily:
To make the shadow of the sunView more efficient, we were telling the view to rasterise, however that caused the label to also rasterise which resulted in blurry text.
Removing rasterisation from the sunView means that the drawing of the shadow is slightly less efficient but that the text is now displayed as desired
Also, updated the colours to use the `withAlphaComponent` function instead of creating a whole new color for standard colours with alpha

#65


fluffyemily:
[This text is low res.](https://trello.com/c/9zBaCy86/177-this-text-is-low-res)

PR #77, opened 2016-11-02
Add non-animating expandable summary views

mcomella:
I tried to get the animations working but it's challenging - let's get this
landed first.

Still TODO (as follow-ups?):
 * Animations (naturally)
 * ScrollView doesn't scroll now that we have the UIPanGestureListener
 * Description separators can sometimes be more than 1px
 * Include expandable drop-down arrow
 * Get yelp logo for description view
 * (if we expect to show places without description views) Make sure view
handles keeping margin below review scores


fluffyemily:
I'm gonna merge this and then take a look at it and see what we can do with the scrolling.

The best thing to do here is probably move the scroll view from the `PageDetailCardViewController` to the `PageDetailViewController`. This means that we won't have two clashing gesture recognizers on the same view. It should also mean that as the view can expand and contract as much as needed without affecting everything else too much.

PR #78, opened 2016-11-02
Issue 11 - place card vertical scrolling

fluffyemily:
Move scroll view to parent view for place card view. 
Set the scroll view content offset through a delegate.
Ensure that vertical scrolling that starts on the current place card doesn't interfere with pan and page horizontal gestures


fluffyemily:

fluffyemily:
@mcomella I've fixed almost everything I think. Underlying scroll view is now handling all vertical scrolling with no interference from us. In order to achieve this, I had to prevent scrolling entirely when the views are unexpanded. This leads to the only remaining niggle:

Currently, when we scroll to an unexpanded view from an expanded view, we shrink the content size and prevent scrolling. This leads to a 'jerk' as the view resets if the scroll is anywhere other than at the top when we pan. However, if I don't reset, then when the expanded view is off screen, it feels odd to have the extra scroll length that leads nowhere. I couldn't decide which to do, so I left it here and thought I would ask your opinion.

PR #79, opened 2016-11-03
Add (some) place details with live data

mcomella:
Not quite done but we could merge this if you'd like.

### DIFF
--- a/Prox/Prox/Models/Place.swift
+++ b/Prox/Prox/Models/Place.swift
@@ -75,7 +75,10 @@ class Place: Hashable {
             return nil
         }
 
-        // TODO: keys to deal with
+        // TODO:
+        // * validate incoming data
+        // * b/c ^, tests
+        // * keys to deal with
         //  - version
         //  - description: utilize provider
         //  - phone
--- a/Prox/Prox/Models/Place.swift
+++ b/Prox/Prox/Models/Place.swift
@@ -109,6 +109,25 @@ enum DayOfWeek: Int {
     case monday = 0 // matches server representation
     case tuesday, wednesday, thursday, friday
     case saturday, sunday
+
+    static func forDate(_ date: Date) -> DayOfWeek {
+        let iOSWeekday = getiOSWeekday(forDate: date)
+        return weekday(fromiOSWeekday: iOSWeekday)
+    }
+
+    private static func getiOSWeekday(forDate date: Date) -> Int {
+        let cal = Calendar(identifier: .gregorian)
+        return cal.component(.weekday, from: date)
+    }
+
+    private static func weekday(fromiOSWeekday iOSWeekday: Int) -> DayOfWeek {
+        // iOSWeekday is 1-7 where Sun = 1; we are 0-6 where Mon = 0
+        var dayInt = iOSWeekday - 2
+        if dayInt < 0 { // only Sunday
+            dayInt = 6
+        }
+        return DayOfWeek(rawValue: dayInt)!
+    }
 }
 
 struct OpenHours {
--- a/Prox/Prox/Models/Place.swift
+++ b/Prox/Prox/Models/Place.swift
@@ -133,4 +133,43 @@ enum DayOfWeek: Int {
 struct OpenHours {
     let startTime: Int
     let endTime: Int
+
+    private static let inputFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "HHmm"
+        return formatter
+    }()
+
+    private static let outputFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.dateStyle = .none
+        formatter.timeStyle = .short // Time will appear in users' clock config: 12hr or 24hr.
+        return formatter
+    }()
+
+    func getStringForStartTime() -> String {
+        return getString(forTime: startTime)
+    }
+
+    func getStringForEndTime() -> String {
+        return getString(forTime: endTime)
+    }
+
+    private func getString(forTime time: Int) -> String {
+        var inputStr = String(time)
+        let len = inputStr.characters.count
+        guard len == 3 || len == 4 else {
+            fatalError("Invalid date str: \(inputStr)")
+        }
+
+        // We expect len 4.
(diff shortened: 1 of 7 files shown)


Prox/Prox/Models/Place.swift
```diff
@@ -109,9 +112,67 @@ enum DayOfWeek: Int {
...
+
+    func getStringForEndTime() -> String {
+        return getString(forTime: endTime)
+    }
+
+    private func getString(forTime time: Int) -> String {
+        var inputStr = String(time)
+        let len = inputStr.characters.count
+        guard len == 3 || len == 4 else {
+            fatalError("Invalid date str: \(inputStr)")
```
fluffyemily: I would prefer not to crash the app if the data from the server is incorrect. Log it, throw an exception, whatever, but don't let the user suffer because we couldn't write the correct unit tests on our server!

Prox/Prox/Models/Place.swift
```diff
@@ -109,9 +112,67 @@ enum DayOfWeek: Int {
...
+        }
+
+        // We expect len 4.
+        if len == 3 {
+            inputStr.insert("0", at: inputStr.startIndex)
+        }
+
+        guard let date = OpenHours.inputFormatter.date(from: inputStr) else {
+            fatalError("Unable to convert input: \(time)")
```
fluffyemily: ditto on the forced crash.

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -186,19 +188,57 @@ class PlaceDetailsCardView: UIScrollView {
         if frame != newFrame { frame = newFrame }
     }
 
+    // TODO: don't duplicate PlaceCarousel code
+    func updateUI(forPlace place: Place) {
```
fluffyemily: In an ideal MVC world the Views should have no knowledge of Models and the Controller should set the view properties, but I am not going to be anal about it here. This is essentially fine as we are not storing the `Place` anywhere in this view.

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -186,19 +188,57 @@ class PlaceDetailsCardView: UIScrollView {
...
+            view.numberOfReviewersLabel.text = "No data found" // TODO: error state.
+            return
+        }
+
+        // TODO: error states
+        view.score = provider.rating ?? 0
+        view.numberOfReviewersLabel.text = "\(provider.totalReviewCount ?? 0) Reviews"
+    }
+
+    private func updateHoursUI(_ hours: [DayOfWeek:OpenHours]?) {
```
fluffyemily: Again, this kind of logic really should be in the `PlaceDetailsCardViewController`, but I can learn to live with it :)

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -186,19 +188,57 @@ class PlaceDetailsCardView: UIScrollView {
...
+        let todayDayOfWeek = DayOfWeek.forDate(Date())
+
+        let primaryText: String
+        let secondaryText: String
+        if hours == nil {
+            primaryText = "Unknown"
+            secondaryText = "No hours data :("
```
fluffyemily: would prefer hours view to be hidden if no hours rather than this text

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -186,19 +188,57 @@ class PlaceDetailsCardView: UIScrollView {
...
+        } else if let todaysHours = hours![todayDayOfWeek] {
+            // TODO: figure out if it's past closing time. or rather filter out place sooner?
+            primaryText = todaysHours.getStringForEndTime()
+            secondaryText = "Closing time"
+        } else {
+            primaryText = "Unknown"
```
fluffyemily: again, I would prefer the hours view to be hidden here, although this text is better than the one above

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -186,19 +188,57 @@ class PlaceDetailsCardView: UIScrollView {
...
+            secondaryText = "Missing today's hours"
+        }
```
fluffyemily: in face, this whole block could be
```
guard let todaysHours = hours?[todayDayOfWeek] else {
    return hoursView.isHidden = true
}

let primaryText = todaysHours.getStringForEndTime()
let secondaryText = "Closing time"
```

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
mcomella: Oops! I'll keep that in mind. I've been thinking in a react-like style (UI state is rendered directly from program state) but perhaps that loses its value when you're mutating the UI, rather than declaring it as a function.

fwiw, I haven't done any MVC beyond the theory so I'm likely to be messy.

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
mcomella: I put this in as a place holder until we had the specs for the error state and Ricardo's error states were text like this, just less... ridiculous. I separated the cases (no hours data vs. no for the current day) as a form of visual debugging.

Though I also see how trying to do the right thing the first time can be faster. In this case, hiding the view breaks the layout so I'm not sure it'd be worth the time to set up.

What do you think? When we're missing specs, is it better to do whatever is easiest or to try to get it right?

mcomella:
After our discussion, we decided to:
* Leave in the fatal errors. While crashing can be a bad user experience, it'll be faster to write the code if we're not handling all the edge cases. Also, not crashing is a better experience if we had analytics to find out where we might display bad data, rather than crash
* Validate the server data when we create the Places, returning nil if the data is invalid. It'd be good to add analytics to identify what the bad data is.
* Error state mocks were released so we don't have to hide the hours view

PR #86, opened 2016-11-04
Add more live data (with hacks for demo)

mcomella:
Some of this is hacked together to meet the demo. Notes on the current state of the app:
* Transit times are not hooked up in the place details – it'll require too much brain for me to break out of the place carousel right now
* There is hacky code to get the description, if it exists, and set it for both description views (else it uses backup text) – the views are not dynamically removed
  * There is also (untested) code to use the new description format, which when the server is updated, should provide a unique description to each view
* The hours code should work but is untested from the server (since data is not in the appropriate format yet and many sites are missing data).
  * Assuming the places don't change, lat: 51.829, lng: -0.215 (White Hart Hotel) should has hours data if we want to present hours for the demo
* Images loaded in the place carousel are buggy – swiping back and forth replaces them with a place holder
* There is no trip advisor data (since it's not on the server)

Everything else is real. :)


Prox/Prox/Models/Place.swift
```diff
@@ -84,62 +117,80 @@ class Place: Hashable {
...
+    }()
 
-    private static func weekday(fromiOSWeekday iOSWeekday: Int) -> DayOfWeek {
-        // iOSWeekday is 1-7 where Sun = 1; we are 0-6 where Mon = 0
-        var dayInt = iOSWeekday - 2
-        if dayInt < 0 { // only Sunday
-            dayInt = 6
-        }
-        return DayOfWeek(rawValue: dayInt)!
+    static func forDate(_ date: Date) -> DayOfWeek {
```
fluffyemily: This is so much cleaner than what was before. Good stuff.

Prox/Prox/Models/Place.swift
```diff
@@ -84,62 +117,80 @@ class Place: Hashable {
...
+     *   - For simplicity, we just use the last open interval for a day.
+     *   - For simplicity, dates are normalized by a single day, allowing comparison but preventing
+     *     the stored dates from being correct (e.g. if today is Mon Nov 7, the stored date for Tues
+     *     isn't Nov 8).
+     *
+     * TODO:
+     *   - Use all open intervals
+     *   - Use accurate days of week in Date (maybe? It adds complexity for little gain).
+     */
+    let hours: [DayOfWeek : (open: Date, close: Date)]
```
fluffyemily: `open` is now a reserved word in Swift 3.0 (like `private` and `public`. I'd be more comfortable if we could find a different name as I am uncertain what the result of this might be long term.

PR #96, opened 2016-11-04
Adding icons

### LINKED ISSUE #95: Implement here, here_small, and map_view icons
`here.png`
`here_small.png`
`map_view icons.png`

vazquez:
Adding `icon_here.png`, `icon_here_small.png`, and modifying `icon_mapview.png`

Fixes #95


mcomella:
This changes the size of the map view icon – is that intentional?

simulator screen shot nov 4 2016 11 46 35

vazquez:
Yes, the shadow was wrong. That was our bad. Although now the icon doesn't look centered :\

mcomella:
> Although now the icon doesn't look centered :\

We could fix it in code (if you provide the specs) or you could just add alpha padding to the asset (this is a prototype after all :).

vazquez:
I updated the icon 👍

PR #109, opened 2016-11-05
Misc. cleanup from Sprint #3 (place details)

mcomella:


Prox/Prox/PlaceDetails/PlaceDetailsCardViewController.swift
```diff
@@ -105,12 +105,13 @@ class PlaceDetailsCardViewController: UIViewController {
...
         setupCardInteractions()
 
-        pageControl.numberOfPages = place.photoURLs?.count ?? 0
+        pageControl.numberOfPages = place.photoURLs.count
     }
 
     private func setLocation(location: CLLocation?) {
         if let lastTravelTimes = place.lastTravelTime {
             self.cardView.updateTravelTimesUI(travelTimes: lastTravelTimes)
+            return
```
fluffyemily: I don't think there should be a return here. What if the user has moved significantly since they last viewed this card. It is possible that the travel times we have cached are now out of date. That is why I left to to display the cached times and then update when the new travel times have been fetched. If I did not do it this way, then it was possible that when scrolling quickly through the cards all the eta's would be spinners.

Possibly what we need to do is add cancelling for travel times, so if a card is no longer visible but the travel times have not completed, we can cancel the request, to stop the requests overloading. I'll have a think about how we could do this and add it to the backlog

mcomella: > It is possible that the travel times we have cached are now out of date. That is why I left to to display the cached times and then update when the new travel times have been fetched.

I'm concerned about showing out-of-date travel times that may arbitrarily update while the user is looking at them. I'd suggest we stop caching but...

> If I did not do it this way, then it was possible that when scrolling quickly through the cards all the eta's would be spinners.

This is an issue that UX would probably agree is important.

> What if the user has moved significantly since they last viewed this card.

The [location manager has functionality](https://developer.apple.com/reference/corelocation/cllocationmanager/1423500-distancefilter) to only update with location events if they exceed a certain distance. (keeping the early return) we could set this filter and, when we receive an event, we could null all cached travel time values, reducing redundant computation (from removing the return value) and providing a better UX.

But then again, it's more complicated to implement so perhaps it's not important enough.

Issue #113, opened 2016-11-07
Current Buddybuild does not use fake location for routing


fluffyemily:
When calculating travel times, the current buddybuild does not use the fake location and so never stops the fetch spinner.

1. Use fake location when fetching current location if fake location required
2. Ensure that spinner is stopped even if travel times returns no value

PR #114, opened 2016-11-07
Issue 113 - Ensure that the fake location used in the buddy build version of the …

fluffyemily:
…app is also used to get directions!

#113


Prox/Prox/PlaceDetails/PlaceDetailsCardViewController.swift
```diff
@@ -111,14 +111,13 @@ class PlaceDetailsCardViewController: UIViewController {
     private func setLocation(location: CLLocation?) {
         if let lastTravelTimes = place.lastTravelTime {
             self.cardView.updateTravelTimesUI(travelTimes: lastTravelTimes)
-            return
```
mcomella: (continuing on the conversation from [the changeset that added this](https://github.com/mozilla-mobile/prox/pull/109#discussion_r86759933)) Let's add a comment explaining why we don't return here.

Issue #103, opened 2016-11-04
Map background image tiles on plus-size devices


mcomella:
See image.

@antlam Do we have an asset for large devices? Or perhaps Emily got that asset herself?

![simulator screen shot nov 4 2016 14 33 19](https://cloud.githubusercontent.com/assets/759372/20023412/169c7cb2-a29d-11e6-976a-a771b843033a.png)

mcomella:
@vazquez, do you have a larger asset for this?

mcomella:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user, I don't want to see UI which is unpolished in edge cases](https://trello.com/c/Fvta5fiT/206-as-a-user-i-don-t-want-to-see-ui-which-is-unpolished-in-edge-cases)

vazquez:
@mcomella - I have updated the map background to include the blur. As for sizes, give these a shot and let me know if they work. I checked the sizes and the 1x 2x and 3x sizes we have should fit all iPhone models.

`map_background.png`
`map_background@2x.png`
`map_background@3x.png`

https://www.dropbox.com/sh/b30k7wfqerbbb7f/AACdQhl1QKr0MnQyly6KGngSa?dl=0

PR #115, opened 2016-11-07
Issue 103 - Don't tile carousel background

fluffyemily:
#103  

@mcomella can you please check to see if it's my eyes playing tricks or whether the artifact is real:

When viewing on a Plus size device (7 Plus or 6x Plus), can you see white horizontal striations across then background just below the carousel?

If they are there and it is not just my mind/screen then this won't do as a solution and we should delete the branch


mcomella:
> can you see white horizontal striations across then background just below the carousel?

I don't, but I wonder if we were seeing different tiling behavior. On my device, your patch fixes #103. Here's before your patch:
simulator screen shot nov 7 2016 13 53 18

And after:
simulator screen shot nov 7 2016 13 51 52

Should we use the larger assets Ricardo added to #103 instead of this solution?

fluffyemily:
Sorry @mcomella I think that is my discalcula kicking in and I saw a 3 as an 8. Sorry for the confusion. I'll update the commit comment

fluffyemily:
I've also updated the image to the larger one. I'll leave this solution because this will scale the image and prevent any tiling regardless of screen size, so it should cope no matter what happens (say, it get's installed on an iPad or something)

Issue #83, opened 2016-11-03
[follow-up] Description view separator unexpectedly expands when view expands


mcomella:
![simulator screen shot nov 3 2016 12 01 56](https://cloud.githubusercontent.com/assets/759372/19981012/585368fe-a1bd-11e6-970e-b3446a0b46da.png)

mcomella:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user, I don't want to see the wikipedia/yelp description separators changing size when I expand the view](https://trello.com/c/GmoP4z5S/207-as-a-user-i-don-t-want-to-see-the-wikipedia-yelp-description-separators-changing-size-when-i-expand-the-view)

PR #117, opened 2016-11-07
Issue 83 - Prevent shadow drawing on expanded Description view

fluffyemily:
#83 

The problem was caused here by the expanding view. The underlying `HorizontalLineView` was essentially being 'stretched' to fit the expanded size, but not being redrawn, causing the line to gert stretched as if it were a shadow.

The solution I implemented here was to move the horizontal line view to the `descriptionTitleView` which does not need to be redrawn when the view expands.

An alternate solution would have been to call `view.setNeedsLayout` on every expansion and contraction but that would have been more resource intensive


Prox/Prox/PlaceDetails/PlaceDetailsDescriptionView.swift
```diff
@@ -129,13 +129,15 @@ class PlaceDetailsDescriptionView: HorizontalLineView {
         expandableLabelHeightConstraint.isActive = false
         expandableLabelBottomConstraint.constant = -10
         logoBottomConstraint.constant = -20
+//        self.setNeedsDisplay()
```
mcomella: rm this.

Prox/Prox/PlaceDetails/PlaceDetailsDescriptionView.swift
```diff
@@ -129,13 +129,15 @@ class PlaceDetailsDescriptionView: HorizontalLineView {
...
         self.layoutIfNeeded()
     }
 
     private func collapseView() {
         expandableLabelBottomConstraint.constant = 0
         expandableLabelHeightConstraint.isActive = true
         logoBottomConstraint.constant = 0
+//        self.setNeedsDisplay()
```
mcomella: rm this

Issue #99, opened 2016-11-04
Implement Event Badge UI


vazquez:
Whenever we send a push notification to a user, we will show them a special card which will contain a badge UI with relevant information to the event.

antlam:
![screen shot 2016-11-04 at 2 04 51 pm](https://cloud.githubusercontent.com/assets/1994820/20022424/bebf83d6-a297-11e6-8761-b3d488e4a1fc.png)

antlam:
![spec_iconevent_padding](https://cloud.githubusercontent.com/assets/1994820/20022595/78a70efe-a298-11e6-8394-1383d81f4848.png)
![spec_event_padding](https://cloud.githubusercontent.com/assets/1994820/20022594/78a2e59a-a298-11e6-8c6d-a111a284aaec.png)
![spec_event_type](https://cloud.githubusercontent.com/assets/1994820/20022596/78a870dc-a298-11e6-8c91-054ddc361701.png)

liuche:
![](https://github.trello.services/images/mini-trello-icon.png) [Add Event badging to the card for events.](https://trello.com/c/bDfP6HSD/196-add-event-badging-to-the-card-for-events)

liuche:
![](https://github.trello.services/images/mini-trello-icon.png) [I want my Prox cards to also include events that will be happening soon because this might make me more interested in visiting the venues.](https://trello.com/c/Zf5kaT3F/195-i-want-my-prox-cards-to-also-include-events-that-will-be-happening-soon-because-this-might-make-me-more-interested-in-visiting-t)

liuche:
@vazquez Question, will the string used in the notification be the same one as the one displayed in the badge? (assuming the "More info" is badge-only)

antlam:
@liuche is this related to what Mcomella pinged about in Slack yesterday? If so, we still aren't sure about the copy yet. Could they be independent?

vazquez:
+1 for the possibility of them being independent

mcomella:
Note: I'm tracking the sub-problems of this issue on the trello card.

PR #139, opened 2016-11-09
Issue #99: Create event badge skeleton for place card view.

mcomella:
The event appears on one place card and is filled with dummy data. No click handlers yet.

Screenshot:
simulator screen shot nov 8 2016 17 34 50


Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -144,17 +152,40 @@ class PlaceDetailsCardView: UIView {
...
     private func setupViews() {
-        backgroundColor = Colors.detailsViewCardBackground
+        backgroundColor = Colors.detailsViewCardBackground // cannot be transparent to display shadow
+        contentView.backgroundColor = Colors.detailsViewCardBackground
+
         layer.cornerRadius = 10
+        contentView.layer.cornerRadius = 10
+        contentView.layer.masksToBounds = true
+
+        addSubview(contentView)
```
fluffyemily: Is there any reason why you didn't just insert the new view into the stack view at the very top?

mcomella:
[I want my Prox cards to also include events that will be happening soon because this might make me more interested in visiting the venues.](https://trello.com/c/Zf5kaT3F/195-i-want-my-prox-cards-to-also-include-events-that-will-be-happening-soon-because-this-might-make-me-more-interested-in-visiting-t)

mcomella:

I didn't use `eventView.removeArrangedSubview` because, according to the docs, this doesn't actually remove the view from the view hierarchy and it keeps displaying (you need to use `UIView.removeFromSuperview` for that) – hiding the view seemed simpler.

fluffyemily:
Not sure why, but this is what I see when I first run through the app. I am wondering whether some of the unsatisfiable constraints are causing this to break.

screen shot 2016-11-10 at 11 55 42

Issue #124, opened 2016-11-08
If past midnight, an open place will be considered closed


mcomella:
Since we get hours by the current day of the week, if it's past midnight, the day of the week is changed but the overnight hours are stored on the previous day. For example:

The overnight hours for Monday are stored on Monday but if it's past midnight, the user will get the hours for Tuesday. Since the store is not indicated as open on Tuesday morning, the user will see the location as closed.

FYI @antlam @vazquez 

Possible solutions:
* Use accurate dates (i.e. dates in 2016) and check for hours on a continuous basis, rather than indexed by day of week
* Check yesterdays hours and if a place is open overnight, take those hours into account
* Duplicate overnight hours to today, morning

PR #140, opened 2016-11-09
Issue #105 - Store opening and closing times as DateComponents

### LINKED ISSUE #105: Open hours logic is not correct (against the server)
This says closed @ 9pm (which means it's after 9pm in my expected logic):

simulator screen shot nov 4 2016 16 27 20

According to Firebase, it should open at 5:30 and close at 9:
screen shot 2016-11-04 at 16 28 06

So the logic is incorrect.

fluffyemily:
Use this to calculate opening and closing time for a given date and provide string helper functions for getting the date as a string.
Display closing time when place is open
Display next opening time when place is closed

#105 
Also addresses #124 

@mcomella I've had a go at solving the open hours quandry. Can you take a peek and let me know what you think?


fluffyemily:
[As a user, I want the open hours in my app correct](https://trello.com/c/gNyrz4SH/201-as-a-user-i-want-the-open-hours-in-my-app-correct)

Prox/Prox/Models/Place.swift
```diff
@@ -237,48 +255,95 @@ struct OpenHours {
...
         return OpenHours(hours: out)
     }
 
-    private static func getDate(fromServerStr serverStr: String) -> Date? {
-        guard serverStr.characters.count != 4 || serverStr.characters.count != 5 else { // "1:00" or "10:00"
-            return nil // the calling function logs.
+    private static func getTimeComponents(fromServerStr serverStr: String) -> DateComponents? {
+        let timeComponents = serverStr.components(separatedBy: ":")
+        guard timeComponents.count == 2,
+            let hours = Int(timeComponents[0]), let minutes = Int(timeComponents[1]) else {
```
mcomella: nit: without the `DateFormatter`, we no longer check if `0 <= hours < 24` and similarly for minutes. I assume this would wrap around for the DateComponent (e.g. 25 hours is day 2, 1 hour) but I'm not sure.

If we ever update the hours format, the DateFormatter would likely be easier to update too.

Prox/Prox/Models/Place.swift
```diff
@@ -237,48 +255,95 @@ struct OpenHours {
...
+            return nil
+        }
+
+        let dateComponents = DateComponents(calendar: OpenHours.calendar, timeZone: nil, era: nil, year: nil, month: nil, day: nil, hour: hours, minute: minutes, second: nil, nanosecond: nil, weekday: nil, weekdayOrdinal: nil, quarter: nil, weekOfMonth: nil, weekOfYear: nil, yearForWeekOfYear: nil)
```
mcomella: nit: there are default values for most of these fields; I don't know if there are advantages to intentionally nilling them though.

Prox/Prox/Models/Place.swift
```diff
@@ -237,48 +255,95 @@ struct OpenHours {
...
+
+        return dateComponents
+    }
+
+    func isOpen(atTime time: Date) -> Bool {
+        let dayOfWeek = DayOfWeek.forDate(time)
+        print("hours \(hours)")
+        print("The day of the week is \(dayOfWeek)")
+        guard let openTime = hours[dayOfWeek]?.openTime,
+            let openingTime = date(forTime: openTime, onDate: time),
```
mcomella: nit: maybe openTime -> openDateComponents? (in the named tuple too) It's a little confusing what the difference between `openTime` and `openingTime` is; same with closing*.

Prox/Prox/Models/Place.swift
```diff
@@ -237,48 +255,95 @@ struct OpenHours {
...
+            var closingTime = date(forTime: closeTime, onDate: time) else {
+            return false
         }
 
-        return inputFormatter.date(from: serverStr)
+        if time < openingTime {
+            print("The previous day of the week is \(dayOfWeek.previousWeekday())")
+            if let yesterdaysClosingHours = hours[dayOfWeek.previousWeekday()]?.closeTime,
+                let yesterdaysClosingTime = date(forTime: yesterdaysClosingHours, onDate: time) {
+                return time < yesterdaysClosingTime
```
mcomella: This only applies if yesterday's closing time < yesterday's opening time.

For example, if we closed yesterday at 11:59pm, this would almost always be true.

Prox/Prox/Models/Place.swift
```diff
@@ -237,48 +255,95 @@ struct OpenHours {
...
+            }
+        }
+
+        if closingTime <= openingTime {
+            closingTime = closingTime.addingTimeInterval(AppConstants.ONE_DAY)
+        }
+
+        return time >= openingTime && time < closingTime
```
mcomella: nit: I had to do a double-take that we intentionally rechecked `time >= openingTime` – should we return from the previous branch and omit the expression down here?

Prox/Prox/Models/Place.swift
```diff
@@ -237,48 +255,95 @@ struct OpenHours {
...
+            let openingTime = date(forTime: openTime, onDate: time) else {
+                return nil
+        }
+
+        if time < openingTime {
+            return timeString(forDate: openingTime)
+        }
+
+        if let tomorrowOpenHours = hours[dayOfWeek.nextWeekday()]?.openTime,
+            let tomorrowOpeningTime = date(forTime: tomorrowOpenHours, onDate: time) {
```
mcomella: nit: probably unnecessary, but we can add 24hrs to `time` just in case we ever supply a date too.

mcomella:
> I think there is one logic error to be fixed though.

Add a test too! :)

Prox/Prox/Models/Place.swift
fluffyemily: Yup, and using a `DateFormatter` will make this bit of code way simpler too. Good point

Prox/Prox/Models/Place.swift
fluffyemily: By changing to use the `DateFormatter` above, this line is removed

fluffyemily:
@mcomella  I've updated the branch according to your suggestions

Prox/Prox/Models/Place.swift
```diff
@@ -282,6 +282,16 @@ struct OpenHours {
         return OpenHours.calendar.dateComponents(timeComponentsSet, from: dateFromTime)
     }
 
+    /** 
+    * The logic for isOpen is as follows:
```
mcomella: nit: I would prefer to use helper functions to make this logic readable in code, rather than having duplication in comments (here & inline). For example:

```
guard ... // return early if no hours today

let openingTimeToday = getDateForOpeningTime(forDate: time, hours: hours)
let closingTimeToday = ... // adds 24hrs if < openingTimeToday
let closingTimeYesterday = ... // subtracts 24hrs b/c yesterday

if let closingTimeYesterday = closingTimeYesterday,
    time < openingTimeToday {
  return time < closingTimeYesterday // We may be open overnight.
}

return time >= openingTimeToday && time < closingTimeToday
```

Prox/ProxTests/OpenHoursTests.swift
```diff
@@ -66,91 +75,71 @@ class OpenHoursTests: XCTestCase {
...
+        date = dateComponents(withYear: 2016, month: 11, day: 7, hour: 7, minute: 0).date!
+        XCTAssertTrue(hours.isOpen(atTime: date), "Expected open for \(date)")
+
+        // current time 7th November 2016 21:00
+        date = dateComponents(withYear: 2016, month: 11, day: 7, hour: 21, minute: 0).date!
+        XCTAssertFalse(hours.isOpen(atTime: date), "Expected closed for \(date)")
+
+        // Monday open time 07:00, closing time 00:00
+        // Tuesday open time 07:00, closing time 02:00
+        let opening = dateComponents(withHour: 7, minute: 0)
```
fluffyemily: Ah! I understand now. Sorry, didn't get that from your previous comment.

Prox/ProxTests/OpenHoursTests.swift
```diff
@@ -66,91 +75,71 @@ class OpenHoursTests: XCTestCase {
...
+        // open time 8th November 2016 07:00
+        // close time 9th November 2016 02:00
+        date = dateComponents(withYear: 2016, month: 11, day: 9, hour: 0, minute: 0).date!
+        XCTAssertTrue(hours.isOpen(atTime: date), "Expected open for \(date)")
+
+        // current time 9th November 2016 02:00
+        // open time 8th November 2016 07:00
+        // close time 9th November 2016 02:00
+        date = dateComponents(withYear: 2016, month: 11, day: 9, hour: 2, minute: 0).date!
+        XCTAssertFalse(hours.isOpen(atTime: date), "Expected closed for \(date)")
```
fluffyemily: I'm pretty sure that we have that test in there already, but I'll check. The fact that you can't see it means that I do need to make the tests clearer

Prox/ProxTests/OpenHoursTests.swift
```diff
@@ -75,53 +75,83 @@ class OpenHoursTests: XCTestCase {
...
 
-        // open time 07:00, closing time 21:00
-        var hours = OpenHours(hours: [.monday : (openTime: dateComponents(withHour: 7, minute: 0), closeTime: dateComponents(withHour: 21, minute: 0))])
+    func testIsClosedWhenCurrentTimeAfterYesterdaysClosingTimeButBeforeTodaysOpeningTime() {
+        let hours = OpenHours(hours: [.sunday : (openTime: dateComponents(withHour: 7, minute: 0), closeTime: dateComponents(withHour: 21, minute: 0)),
+                                      .monday : (openTime: dateComponents(withHour: 7, minute: 0), closeTime: dateComponents(withHour: 21, minute: 0))])
+        let date = monday(atHour: 6, minute: 59)
 
-        XCTAssertFalse(hours.isOpen(atTime: date), "Expected closed for \(date)")
+        isClosed(openHours: hours, time: date)
```
mcomella: nit: -> `assertIsClosed`. Same for `isOpen`

Prox/ProxTests/OpenHoursTests.swift
```diff
@@ -75,53 +75,83 @@ class OpenHoursTests: XCTestCase {
...
+    }
 
-        // current time 7th November 2016 07:00
-        date = dateComponents(withYear: 2016, month: 11, day: 7, hour: 7, minute: 0).date!
-        XCTAssertTrue(hours.isOpen(atTime: date), "Expected open for \(date)")
+    func testIfOpenWhenCurrentTimeIsTodaysOpeningTime() {
```
mcomella: nit: `Is` not `If`

Prox/Prox/Models/Place.swift
```diff
@@ -282,6 +282,29 @@ struct OpenHours {
...
+    }
+
+    private func getClosingTime(forDate date: Date) -> Date? {
+        guard let openingTime = getOpeningTime(forDate: date),
+            let closeDateComponents = hours[DayOfWeek.forDate(date)]?.closeTime,
+            var closingTime = getDate(forTime: closeDateComponents, onDate: date) else {
+                return nil
+        }
+
+        // Add 24 hours to the closing time if the closing time if before the opening time
```
mcomella: nit: I'd explain why we're doing this (since it's open overnight)

Prox/Prox/Models/Place.swift
```diff
@@ -237,48 +262,113 @@ struct OpenHours {
...
+        guard let openingTime = getOpeningTime(forDate: time),
+            let closingTime = getClosingTime(forDate: time) else {
+            return false
         }
 
-        return inputFormatter.date(from: serverStr)
+        // Is the current time before the opening time and the venue closed after midnight the previous day?
+        if time < openingTime,
+            let yesterdaysClosingTime = getClosingTime(forDate: time.addingTimeInterval(-AppConstants.ONE_DAY)),
+            yesterdaysClosingTime < openingTime {
```
mcomella: Why is `yesterdaysClosingTime < openingTime` necessary?

PR #144, opened 2016-11-09
Closes #85: Match updated mocks for "You're Here" UI.

### LINKED ISSUE #85: "Current Location" UI When User is at Location
When the user is in the venue's location, we don't need to show the Location UI. For these cases, we need to hide the "Current Location" UI and show a "You're here!" Badge.

mcomella:
#85

Post-patch screenshot:
simulator screen shot nov 9 2016 15 36 25

### DIFF
--- a/Prox/Prox/PlaceDetails/PlaceDetailsIconInfoView.swift
+++ b/Prox/Prox/PlaceDetails/PlaceDetailsIconInfoView.swift
@@ -32,12 +32,21 @@ class PlaceDetailsIconInfoView: UIView {
         return view
     }()
 
-
     lazy var loadingSpinner: UIActivityIndicatorView = {
         let indicatorView = UIActivityIndicatorView(activityIndicatorStyle: UIActivityIndicatorViewStyle.gray)
         indicatorView.hidesWhenStopped = true
         return indicatorView
     }()
+
+    // We hide primary because we want secondary text style.
+    var isPrimaryTextLabelHidden = false {
+        didSet {
+            if oldValue != isPrimaryTextLabelHidden { setPrimaryTextLabelHidden(isPrimaryTextLabelHidden) }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


Prox/Prox/PlaceDetails/PlaceDetailsIconInfoView.swift
```diff
@@ -70,18 +79,33 @@ class PlaceDetailsIconInfoView: UIView {
     }
 
     private func setupLabelContainerSubviews() -> [NSLayoutConstraint] {
-        labelContainer.addSubview(primaryTextLabel)
-        var constraints = [primaryTextLabel.topAnchor.constraint(equalTo: labelContainer.topAnchor),
-                           primaryTextLabel.leadingAnchor.constraint(equalTo: labelContainer.leadingAnchor),
-                           primaryTextLabel.trailingAnchor.constraint(equalTo: labelContainer.trailingAnchor),
-                           primaryTextLabel.heightAnchor.constraint(equalTo: labelContainer.heightAnchor, multiplier: 0.5)]
+        secondaryTextLabelHalfHeightConstraint = secondaryTextLabel.heightAnchor.constraint(equalTo: labelContainer.heightAnchor, multiplier: 0.5)
```
fluffyemily: You don't need 2 different constraints here, you just need the reference to a single constraint with which you can change the `multiplier` value as required

fluffyemily: If you put the 2 different multipliers as constants then you can you just refer to the constants when setting them

mcomella: Unlike  `NSLayoutConstraint.constant`, `multiplier` is a getter only. :( I'm going to leave it as is and merge.

PR #147, opened 2016-11-10
Closes #84: Add forward arrow to directions view

### LINKED ISSUE #84: Implement forward-arrow on Summary Card
We received feedback from the User Testing round that people did not know where to tap to receive walking (or driving) directions. As such, we are modifying the design just a bit.

We are using the iOS' `arrow-forward` icon. Exported assets can be found here: https://www.dropbox.com/sh/b30k7wfqerbbb7f/AACdQhl1QKr0MnQyly6KGngSa?dl=0

`icon_forwardarrow.png`
`icon_forwardarrow@2x.png`
`icon_forwardarrow@3x.png`

screenshot 2016-11-03 15 22 34

screenshot 2016-11-03 15 22 19

mcomella:
(Sorry, includes previous changes but github doesn't let me make the PR against my own branch since it's in my fork 🐳)

#84. Still TBD in bug: expected height of the forward arrow asset.


Prox/Prox/PlaceDetails/PlaceDetailsIconInfoView.swift
```diff
@@ -32,15 +32,29 @@ class PlaceDetailsIconInfoView: UIView {
         return view
     }()
 
+    lazy var forwardArrowView = UIImageView(image: UIImage(named: "icon_forward"))
```
fluffyemily: I would prefer it is the `forwardArrowView` was a generic `accessoryItemView` or something, and the `PlaceDetailsCardView` sets the image depending on whether or not it is required, rather than the image being fixed inside the `PlaceDetailsIconInfoView` (like we are ding for the `iconView`. Doing it in the `PlaceDetailsIconInfoView` feels like it removes the generic nature of the view itself.

Also, it would remove the need to hide or show the `forwardArrowView`. It either has an icon or it doesn't.

mcomella: This is a good idea. However, to save time I'll leave it as is and make it generic if we need to in the future.

PR #148, opened 2016-11-10
Closes #136 - add background gradient

### LINKED ISSUE #136: Background image Gradient on Summary Cards
Based on feedback for User Research, the map background we implemented in Summary Cards confused individuals, as we offered no actionable interaction.

We are changing the background image on Summary Cards to offer a blurred image with a gradient.

## Gradient Specs

home - a bay

mcomella:
#136. Still TBD in bug – apply gradient before or after blur effect.


mcomella:
Confirmed in #136 that this is what we want.

PR #149, opened 2016-11-10
Card 210 – Pinging the server and waiting for a response.

jhugman:
https://trello.com/c/rZfCz8qI/210-as-a-user-i-want-to-see-places-cards-from-the-area-that-i-m-in-so-i-can-explore-more-than-just-hawaii

This is the first story in roundtripping with the server.

This pull request adds a new PlacesController (and includes a PlacesControllerDelegate) for telling the server where we are and subsequently waiting for the server to come back with some data.

It includes a very rudimentary mechanism (polling) to check if firebase has the complete set of venues available for this location.

It does not include:
 
 * [A UX approved UI to wait for results](https://trello.com/c/Y1cwkXeM/163-as-a-user-i-want-to-have-an-enjoyable-experience-in-the-time-it-takes-to-discover-the-places-around-me-so-that-i-do-not-quit-the)
 * An incremental load of place, which would ultimately be more satisfying for the user and efficient of network usage.

### DIFF
--- a/Prox/Prox/Data/PlacesController.swift
+++ b/Prox/Prox/Data/PlacesController.swift
@@ -1,0 +1,115 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import AFNetworking
+import Foundation
+import Deferred
+
+/*
+ * Delegate methods for updating places asynchronously.
+ * All methods on the delegate will be called on the main thread.
+ */
+protocol PlacesControllerDelegate: class {
+    func placeControllerWillStartFetchingPlaces(_ controller: PlacesController)
+    func placesController(_ controller: PlacesController, didReceivePlaces places: [Place])
+    func placeControllerDidFinishFetchingPlaces(_ controller: PlacesController)
+    func placesController(_ controller: PlacesController, didError error: Error)
+}
+
+private let apiSuffix = "/api/v1.0/at/%f/%f"
+private let numberOfRetries = 60
+private let timeBetweenRetries = 1
+
+class PlacesController {
+    weak var delegate: PlacesControllerDelegate?
+
+    private let database = FirebasePlacesDatabase()
+
+    private var isUpdating = false
+
+    private lazy var sessionManager: AFHTTPSessionManager = {
+        let manager = AFHTTPSessionManager(sessionConfiguration: .default)
+        manager.responseSerializer = AFHTTPResponseSerializer()
+        return manager
+    }()
+
+    func updatePlaces(forLocation location: CLLocation) {
+        assert(Thread.isMainThread)
+        // We would like to prevent running more than one update at the same time.
+        if isUpdating {
+            return
+        }
+        isUpdating = true
+
+        self.delegate?.placeControllerWillStartFetchingPlaces(self)
+
+        // Tell the server that we're here. We don't need to do anything with the return
+        // value: we're just instructing the server to go crawl the venues/events the area 
+        // we're in now.
+        let coord = location.coordinate
+        let path  = String(format:apiSuffix, coord.latitude, coord.longitude)
+        let url   = AppConstants.serverURL.appendingPathComponent(path)
+
+        sessionManager.put(url.absoluteString, parameters: nil,
+            success: { (task, data) in
+                // TODO is there anything useful we can get from the server to help the UI here?
+                // e.g. some JSON to give queue size (how busy is it).
+                print("Server responded ok.")
+            },
+            failure: { (task, err) in
+                print("Error from server: \(err)")
+                DispatchQueue.main.async {
+                    self.delegate?.placesController(self, didError: err)
+                }
+            }
+        )
+
+        // Immediately after we've told the server, we should start querying the firebase datastore,
+        // because we may have already got something cached.
+        retryQueryPlaces(location: location, retriesLeft: numberOfRetries, lastCount: 0)
+    }
+
+    private func retryQueryPlaces(location: CLLocation, retriesLeft: Int, lastCount: Int) {
+        // Fetch a stable list of places from firebase.
+        // In the event of the server crawling (from a cold start, for example)
+        // the server will be adding places to firebase.
+        // We want to wait for the number of firebase results to stop changing.
+        database.getPlaces(forLocation: location).upon { results in
+            let places = results.flatMap { $0.successResult() }
+            let placeCount = places.count
+            // Check if we have a stable number of places.
+            if (placeCount > 0 && lastCount == placeCount) || retriesLeft == 0 {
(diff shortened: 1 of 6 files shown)


jhugman:
[As a user, I want to see places cards from the area that I'm in so I can explore more than just Hawaii](https://trello.com/c/rZfCz8qI/210-as-a-user-i-want-to-see-places-cards-from-the-area-that-i-m-in-so-i-can-explore-more-than-just-hawaii)

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -207,7 +212,7 @@ class PlaceCarouselViewController: UIViewController {
         // this ensures consistency between the place details and the carousel underneath
         // and makes sure we don't end up providing weird data to users while they are scrolling
         // through places details
-        updatePlaces(forLocation: location)
+        updatePlaces(for: location)
```
fluffyemily: nit: Not sure I like this arg name change. What if we'd decided to call `location` `fred`? Suddenly the meaning derived from the parameter name is lost and the function call is less explicit.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -360,3 +358,26 @@ extension PlaceCarouselViewController: LocationProvider {
...
+            // TODO placeholder for the waiting state.
+            headerView.numberOfPlacesLabel.text = "Waiting"
+        } else {
+            self.places = []
+        }
+
+    }
+
+    func placesController(_ controller: PlacesController, didUpdatePlaces places: [Place]) {
+        self.places = places
```
fluffyemily: can we please have the full functionality from the removed closure on the old `getPlaces` request?
```
if places != self.places {
    self.places = places
    self.placeCarousel.refresh()
}
```

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -360,3 +358,26 @@ extension PlaceCarouselViewController: LocationProvider {
     }
 }
 
+extension PlaceCarouselViewController: PlacesControllerDelegate {
+    func placesControllerNoPlacesFound(_ controller: PlacesController, willRetry retry: Bool) {
+        if retry {
+            // TODO placeholder for the waiting state.
+            headerView.numberOfPlacesLabel.text = "Waiting"
+        } else {
+            self.places = []
```
fluffyemily: add `self.placeCarousel.refresh()` else the placeCarousel will never update to the new value

Prox/Prox/Data/PlacesController.swift
```diff
@@ -0,0 +1,83 @@
...
+ * Delegate methods for updating places asynchronously.
+ * All methods on the delegate will be called on the main thread.
+ */
+protocol PlacesControllerDelegate: class {
+    func placesController(_ controller: PlacesController, didUpdatePlaces places: [Place])
+    func placesControllerNoPlacesFound(_ controller: PlacesController, willRetry retry: Bool)
+    func placesController(_ controller: PlacesController, didError error: Error)
+}
+
+let apiSuffix = "/api/v1.0/at/%f/%f"
```
fluffyemily: as this is not used anywhere outside of `PlacesController` this should be marked as `private`

Prox/Prox/Data/PlacesController.swift
```diff
@@ -0,0 +1,83 @@
...
+
+class PlacesController {
+    weak var delegate: PlacesControllerDelegate?
+
+    private let database = FirebasePlacesDatabase()
+
+    func updatePlaces(for location: CLLocation) {
+        let sessionManager = AFHTTPSessionManager(sessionConfiguration: .default)
```
jhugman: ok

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
jhugman: `self.placeCarousel.refresh()` is called at https://github.com/mozilla-mobile/prox/pull/149/files#diff-28319a278e726e589796867f78f1bfc0R57

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
jhugman: The `refresh` is redundant (see above).

I've put the inequality test in the setter, to make it idempotent.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -0,0 +1,104 @@
...
+                print("Error from server: \(err)")
+                DispatchQueue.main.async {
+                    self.delegate?.placesController(self, didError: err)
+                }
+            }
+        )
+
+        // Immediately after we've told the server, we should start querying the firebase datastore,
+        // because we may have already got something cached.
+        retryQueryPlaces(location: location, retriesLeft: 60, lastCount: 0)
```
fluffyemily: nit: would be nice if `retriesLeft` param was a constant

Prox/Prox/Data/PlacesController.swift
```diff
@@ -0,0 +1,104 @@
...
+        // the server will be adding places to firebase.
+        // We want to wait for the number of firebase results to stop changing.
+        database.getPlaces(forLocation: location).upon { results in
+            let places = results.flatMap { $0.successResult() }
+            let placeCount = places.count
+            // Check if we have a stable number of places.
+            if (placeCount > 0 && lastCount == placeCount) || retriesLeft == 0 {
+                let sortedPlaces = self.preparePlaces(places: places, forLocation: location)
+                DispatchQueue.main.async {
+                    self.delegate?.placesController(self, didReceivePlaces: sortedPlaces)
```
fluffyemily: it would be nice if this happened in every recursion, rather than just at the end. That way we can start displaying places as soon as we start getting them, rather than only when we have fetched them all. It would make the wait time for full results less obvious. But maybe that is an enhancement for later

Prox/Prox/Data/PlacesController.swift
```diff
@@ -0,0 +1,104 @@
...
+protocol PlacesControllerDelegate: class {
+    func placeControllerWillStartFetchingPlaces(_ controller: PlacesController)
+    func placesController(_ controller: PlacesController, didReceivePlaces places: [Place])
+    func placeControllerDidFinishFetchingPlaces(_ controller: PlacesController)
+    func placesController(_ controller: PlacesController, didError error: Error)
+}
+
+private let apiSuffix = "/api/v1.0/at/%f/%f"
+
+class PlacesController {
```
fluffyemily: Based on a comment that @mcomella made on another patch where I was trying to copy this pattern, perhaps a different name approach would be `PlacesProvider` so that we don't get confused between ViewControllers nd Model Controllers?

jhugman:
Added changes.

jhugman:
Squashed and merged.

Issue #125, opened 2016-11-08
Detect when the user has been at a location for at least 15 minutes


fluffyemily:
We need to check for events near the users current location but only when the user has been at that location for at least 15 minutes.

Set up a small region around the users current location to monitor when they exit that region. 
Set a timer to fire in 15 minutes
If the user exits the region before the time is fired, cancel the timer
If the user does not exit the region before the timer is fired, search for events matching our event criteria for that location
Set a timer to fire every hour that the user does not move from that location

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user I want the app to be aware when my location changes even when I am not using the app so that I can be informed when new places and events are near me as I move about](https://trello.com/c/NESzoTGn/139-as-a-user-i-want-the-app-to-be-aware-when-my-location-changes-even-when-i-am-not-using-the-app-so-that-i-can-be-informed-when-ne)

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user I want to be notified when a relevant event is starting soon in a location near me so that I can be delighted by things that I would not have thought to look for](https://trello.com/c/XCFZIfuk/108-as-a-user-i-want-to-be-notified-when-a-relevant-event-is-starting-soon-in-a-location-near-me-so-that-i-can-be-delighted-by-thing)

PR #150, opened 2016-11-10
Issue 125 - detect when user in location more than 15 mins

fluffyemily:
#125

This is the first part of https://trello.com/c/XCFZIfuk/108-as-a-user-i-want-to-be-notified-when-a-relevant-event-is-starting-soon-in-a-location-near-me-so-that-i-can-be-delighted-by-thing


fluffyemily:
[As a user I want to be notified when a relevant event is starting soon in a location near me so that I can be delighted by things that I would not have thought to look for](https://trello.com/c/XCFZIfuk/108-as-a-user-i-want-to-be-notified-when-a-relevant-event-is-starting-soon-in-a-location-near-me-so-that-i-can-be-delighted-by-thing)

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -101,6 +114,17 @@ class PlaceCarouselViewController: UIViewController {
...
     fileprivate var fakeLocation: CLLocation = CLLocation(latitude: 19.9263136, longitude: -155.8868328)
 
+    fileprivate var monitoredRegions: [String: GeofenceRegion] = [String: GeofenceRegion]()
+    fileprivate var shouldFetchEvents: Bool {
+        guard var lastLocationFetchTime = timeOfLastLocationUpdate else {
+            return false
+        }
+        let now = Date()
+        lastLocationFetchTime += minimumTimeAtLocationBeforeFetchingEvents
+        return lastLocationFetchTime < now
```
mcomella: nit: it's a little confusing that `lastLocationFetchTime` is actually `timeToFetchEvents` here (since we mutate it). I'd make a new var to store the incremented value with a name like ^.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -31,8 +31,20 @@ struct PlaceDataSourceError: Error {
 
 class PlaceCarouselViewController: UIViewController {
 
+    fileprivate let minimumTimeAtLocationBeforeFetchingEvents: TimeInterval = 15 * 60
```
mcomella: nit: -> `minimumIntervalAtLocation...`

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -254,6 +311,53 @@ class PlaceCarouselViewController: UIViewController {
...
+                }
+            }
+        #else
+            sunriseSet = EDSunriseSet(timezone: NSTimeZone.local, latitude: coord.latitude, longitude: coord.longitude)
+        #endif
+
+    }
+
+    fileprivate func startMonitoring(location: CLLocation, withIdentifier identifier: String, forEntry: (()->())?, forExit: (()->())?) {
+        let region = GeofenceRegion(location: location.coordinate, identifier: identifier, radius: 50.00, onEntry: forEntry, onExit: forExit)
```
mcomella: nit: radius as constant

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -31,8 +31,20 @@ struct PlaceDataSourceError: Error {
...
+    fileprivate let minimumTimeAtLocationBeforeFetchingEvents: TimeInterval = 15 * 60
+    fileprivate let currentLocationIdentifier = "CURRENT_LOCATION"
     fileprivate let MIN_SECS_BETWEEN_LOCATION_UPDATES: TimeInterval = 1
-    fileprivate var timeOfLastLocationUpdate: Date?
+
+    private let timeOfLastLocationUpdateKey = "timeOfLastLocationUpdate"
+
+    fileprivate var timeOfLastLocationUpdate: Date? {
+        get {
+            return UserDefaults.standard.value(forKey: timeOfLastLocationUpdateKey) as? Date
```
mcomella: On a fresh install, I think `timeOfLastLocationUpdate` will always be nil.

Also, I can't seem to find the `UserDefaults.value` method – the one I found was [`object`](https://developer.apple.com/reference/foundation/userdefaults/1410095-object). Know what might be going on there?

mcomella:
@fluffyemily Question: we have to show events on place cards so why do we need to fetch events every 15 minutes when a user stayed at a location for long enough? Is this to save battery/network when running in the background (as opposed to getting events every minimum duration)?

fluffyemily:
This is for event notifications. If we don't keep checking for events if we
have been in a place for at least 15 minutes then we won't be able to post
a notification. Without this kind of background polling, we will only see
events when we are in the app, not when we are outside of the app.

The reason we keep polling when we are in the foreground is that we need to
alert the user to the fact that an event has been detected even when they
are in the app, especially as we don't update places if the user is in the
details screen. Therefore if an event is discovered on a place that is
currently not in our filtered list, we cannot update that list until the
user has closed the app or closed the detail view. If they don't do this
then the user may never know about the event.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: yes, that is the point. This allows our check for `timeOfLastLocationUpdate` in `locationManager(_:didUpdateLocations:)` to still work, so the very first time we start the app, when we have never fetched a location before, we can process the new location and set a value here for the first time.

`UserDefaults.standard.value(forKey:)` is there because `UserDefaults` is backed by a hashtable and inherits that function through there. I'll update to use `UserDefaults.standard.object(forKey:)` to make it use the documented part of the api (the old apple docs used to give you links to the inheritance tree, but I can't see them in the updated docs)

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: I'll add a helper function

Issue #151, opened 2016-11-10
Search for events when user has been in current location for more than 15 minutes


fluffyemily:
We currently call the `fetchEvents` function whenever the user has been in the current location for more than 15 minutes, but this function does nothing. We need to query the events DB to see whether there are any events that match our event criteria available

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user I want to be notified when a relevant event is starting soon in a location near me so that I can be delighted by things that I would not have thought to look for](https://trello.com/c/XCFZIfuk/108-as-a-user-i-want-to-be-notified-when-a-relevant-event-is-starting-soon-in-a-location-near-me-so-that-i-can-be-delighted-by-thing)

PR #156, opened 2016-11-10
Issue 151 - fetch events

fluffyemily:
#151 

Uses a FakeEventsDatabase implementation to provide fake events for now

Uses `EventsController` pattern as created by @jhugman in https://github.com/mozilla-mobile/prox/pull/149 so that the `PlaceCarouselViewController` and the `EventManager` don't have to care about promises/futures

Creates an `EventManager` to handle events so that `PlaceCarouselViewController` doesn't get too bloated with functionality that is not directly related to the display of events.

I have a horrible feeling that there is something really major that I have totally overlooked here, but I cannot for the life of me figure out what it is. If you spot something, please let me know.


fluffyemily:
[As a user I want to be notified when a relevant event is starting soon in a location near me so that I can be delighted by things that I would not have thought to look for](https://trello.com/c/XCFZIfuk/108-as-a-user-i-want-to-be-notified-when-a-relevant-event-is-starting-soon-in-a-location-near-me-so-that-i-can-be-delighted-by-thing)

Prox/Prox.xcodeproj/project.pbxproj
```diff
@@ -488,7 +503,7 @@
 				TargetAttributes = {
 					7B29E7BE1DA3EC360099AF38 = {
 						CreatedOnToolsVersion = 8.0;
-						DevelopmentTeam = 9G8J6YA743;
```
mcomella: Development team changed.

Did you know you can add partial files to the staging area with `git add -p <file-name>`? I'm not sure why the dev team changes but you could selective add the project file if this is a common problem.

Prox/Prox/AppDelegate.swift
```diff
@@ -75,7 +75,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func application(_ application: UIApplication, performFetchWithCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
```
mcomella: How do we activate this? Does it happen automatically with the `*.plist` category?

Prox/Prox/Data/EventsController.swift
```diff
@@ -0,0 +1,25 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+protocol EventsControllerDelegate: class {
+    func eventController(_ eventController: EventsController, didUpdateEvents: [Event])
+    func eventController(_ eventController: EventsController, didError error: Error)
+}
+
+class EventsController {
```
mcomella: I think there is  an extra unnecessary level of indirection. EventsManager -> EventsController -> EventsDatabase. Why can't EventsManager or EventsController talk to EventsDatabase directly?

nit: `*Controller` clashes with `ViewController` in my mind – I wonder if there's a better name.

Prox/Prox/Utilities/AppConstants.swift
```diff
@@ -13,7 +13,9 @@ public enum AppBuildChannel {
 public struct AppConstants {
 
     public static let isRunningTest = NSClassFromString("XCTestCase") != nil
-    public static let backgroundFetchInterval: TimeInterval = 5 * 60
+    public static let backgroundFetchInterval: TimeInterval = 1 * 60
+    public static let minimumTimeAtLocationBeforeFetchingEvents: TimeInterval = 1 * 60
```
mcomella: Do we want to commit 60s updates for all builds? Perhaps we should separate this depending on local, BB, etc.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -31,21 +31,20 @@ struct PlaceDataSourceError: Error {
 
 class PlaceCarouselViewController: UIViewController {
 
-    fileprivate let minimumTimeAtLocationBeforeFetchingEvents: TimeInterval = 15 * 60
     fileprivate let currentLocationIdentifier = "CURRENT_LOCATION"
     fileprivate let MIN_SECS_BETWEEN_LOCATION_UPDATES: TimeInterval = 1
 
-    private let timeOfLastLocationUpdateKey = "timeOfLastLocationUpdate"
-
     fileprivate var timeOfLastLocationUpdate: Date? {
```
mcomella: I think this is unused now.

Prox/Prox/Utilities/EventManager.swift
```diff
@@ -0,0 +1,50 @@
...
+        guard var lastLocationFetchTime = timeOfLastLocationUpdate else {
+            return false
+        }
+        let now = Date()
+        lastLocationFetchTime += AppConstants.minimumTimeAtLocationBeforeFetchingEvents
+        return lastLocationFetchTime < now
+    }
+
+    fileprivate var timeOfLastLocationUpdate: Date? {
+        return UserDefaults.standard.value(forKey: AppConstants.timeOfLastLocationUpdateKey) as? Date
```
mcomella: There's no longer a setter here and, as per the previous PR, I don't think it's ever set.

Prox/Prox/Data/EventsController.swift
```diff
@@ -0,0 +1,25 @@
...
+    lazy var eventsDatabase: EventsDatabase = FakeEventsDatabase()
+
+    weak var delegate: EventsControllerDelegate?
```
mcomella: Is the advantage of using the `Controller` with a `Delegate` here (rather than having the delegate set directly on the Database) the same as in an MVC? Specifically, creating an interface between storage and the application logic, making the code more modular for testing and re-use?

Prox/Prox/Utilities/AppConstants.swift
fluffyemily: Yup, that is a mistake. I forgot to swap the values back to 5 & 15 minutes from when I was testing (I didn't want to wait that long for a background fetch to trigger)

Prox/Prox/Data/EventsController.swift
fluffyemily: You're right with naming. I'll rename `EventsManager` to `EventNotificationManager` or something. I only used `EventController` because that was the name that @jhugman chose for his version of this for places and I wanted consistency in naming.

Prox/Prox/Data/EventsController.swift
fluffyemily: This will become clearer with future patches as I start sending notifications. Will rename to `EventNotificationManager` to make it clearer.

Prox/Prox.xcodeproj/project.pbxproj
fluffyemily: This happens when I deploy to device using the Prox Enterprise scheme. I keep forgetting to check for it when I check stuff in. It is part of Apple's automatic codesigning thing that is a very new thing, but essential for the smooth running of CI.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: Nope, this is still used as we still set `timeOfLastLocationUpdate` key in this class (for as long as it is before I move the location handling out of `PlaceCarouselViewController`

PR #166, opened 2016-11-14
Closes #161 - prevent completion handler from being called twice with…

### LINKED ISSUE #161: [Crash] NSInternalInconsistencyException: you can't call -sendResponse: twice nor after encoding it [Build #247]
Buddybuild detected a crash from mpopova@mozilla.com
`Prox/AppDelegate.swift` line 9`
```swift
import UIKit
import Firebase

@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate { <-- CRASHED

    var window: UIWindow?
    var placeCarouselViewController: PlaceCarouselViewController?

    private var authorizedUser: FIRUser?
```

[View File](https://github.com/mozilla-mobile/prox/blob/f3bd1ae87e473385d3721c71ba6defd6e39a9126/Prox/AppDelegate.swift#L9) | [View Full Crash Details](https://dashboard.buddybuild.com/apps/57f51f8bcaa70a01005412ee/crashreport/27b13f33f4104e50b74674421bd87158)

mcomella:
… unique callbacks.

NOTE: I was unable to reproduce the crash and, in any case, don't know how to
test this code so I don't know if this still works as expected - I fixed it
based on continued compilation.

The crash was caused because the delegate contained a mutable completion
handler which, if `fetchEvents` was called more than once, would be replaced,
having the same system completion handler called multiple times.

I removed the delegate in favor of a callback, which seemed simpler. If we
wanted to continue to use the delegate, we could pass a unique identifier into
the fetchEvents call and return that to the delegate, keeping a list of IDs to
completion handlers, but that seemed complicated.

@fluffyemily PLEASE test this when you review! :)

### DIFF
--- a/Prox/Prox/Data/EventsController.swift
+++ b/Prox/Prox/Data/EventsController.swift
@@ -12,13 +12,11 @@ protocol EventsControllerDelegate: class {
 class EventsController {
     lazy var eventsDatabase: EventsDatabase = FakeEventsDatabase()
 
-    weak var delegate: EventsControllerDelegate?
-
-    func getEvents(forLocation location: CLLocation) {
+    func getEvents(forLocation location: CLLocation, completion: @escaping (([Event]?, Error?) -> Void)) {
         return eventsDatabase.getEvents(forLocation: location).upon { results in
             let events = results.flatMap { $0.successResult() }
             DispatchQueue.main.async {
-                self.delegate?.eventController(self, didUpdateEvents: events)
+                completion(events, nil)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


fluffyemily:
The crash was caused during a background fetch. The reason the stack trace was pretty useless was because the app was being awoken in the background and therefore wasn't actually running. Also, iOS stack traces can be a little crap.

To test a background fetch, you can do the following 

1. Edit the scheme you are currently running (`Product -> Scheme -> Edit Scheme`). 
2. Under `Run` in the `Options` tab, check the `Launch due to background fetch event` option and `Close` the dialog. 
3. Run the app. This time it will not open with the UI, but `AppDelegate.application(_ application:, performFetchWithCompletionHandler)` will be called instead, just as it would do for a background fetch event in real life.

PR #173, opened 2016-11-15
Add RemoteConfig (remotely changeable configuration).

jhugman:
This adds three commits:

1. Firebase's RemoteConfig library, which is updated from the remote server. There is a pattern to add keys, defaults and values. The debug expiry time is 60 seconds.
2. Values for search radius that the app will query firebase for venues.
3. Maximum walking time to use before the user is told to drive.

From here on in, I'd recommend using this for any magic numbers, or strings, or arrays that will materially affect the business logic of the app.

### DIFF
--- a/Prox/Prox/AppDelegate.swift
+++ b/Prox/Prox/AppDelegate.swift
@@ -4,6 +4,7 @@
 
 import UIKit
 import Firebase
+import FirebaseRemoteConfig
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
@@ -13,10 +14,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     private var authorizedUser: FIRUser?
 
+    private lazy var remoteConfigCacheExpiration: TimeInterval = {
+        if AppConstants.isDebug {
+            // Refresh the config if it hasn't been refreshed in 60 seconds.
+            return 60
+        }
+        let key = RemoteConfigKeys.remoteConfigCacheExpiration
+        return FIRRemoteConfig.remoteConfig()[key].numberValue!.doubleValue
+    }()
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
-
         setupFirebase()
+        setupRemoteConfig()
         BuddyBuildSDK.setup()
         application.setMinimumBackgroundFetchInterval(AppConstants.backgroundFetchInterval)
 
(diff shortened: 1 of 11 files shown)


fluffyemily:
Not sure why, but I had to add `RemoteConfigKeys.swift` to the ProxTests target before tests would compile, complaining that it couldn't find the file.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -34,6 +35,11 @@ class PlacesController {
         return manager
     }()
 
+    private lazy var radius: Double = {
+        let key = RemoteConfigKeys.searchRadiusInKm
+        return FIRRemoteConfig.remoteConfig()[key].numberValue!.doubleValue
```
fluffyemily: are we guaranteed to have a value here?

jhugman:
Yes. There is a value in the `RemoteConfigDefaults.plist`.

The only thing to worry about here: what happens if we fat finger a setting
in the RemoteConfig console, such that we don't have a valid number here.

Mitigating this would mean moving the defaults out of the plist and into
code, and about a screenful of more code.

i.e.
```swift
class RemoteConfigKeys {
    public let searchRadiusInKm =
RemoteConfigValueDouble(key:"search_radius_in_km", defaultValue: 4)
}
```

```swift
let radius = RemoteConfigKeys.searchRadiusInKm.value
```

This would be nicer, but I wanted to land a simple thing before landing the
more engineered version.

Prox/Prox/Data/PlacesController.swift
jhugman: Yes. There is a value in the `RemoteConfigDefaults.plist`.

The only thing to worry about here: what happens if we fat finger a setting
in the RemoteConfig console, such that we don't have a valid number here.

Mitigating this would mean moving the defaults out of the plist and into
code, and about a screenful of more code.

e.g.

```swift
class RemoteConfigKeys {
    public let searchRadiusInKm =
RemoteConfigValueDouble(key:"search_radius_in_km", defaultValue: 4)
}
```

Then when we use it:

```swift
let radius = RemoteConfigKeys.searchRadiusInKm.value
```

This would be nicer, but I wanted to land a simple thing before landing the
more engineered version.

Prox/Prox/AppDelegate.swift
```diff
@@ -13,10 +14,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     private var authorizedUser: FIRUser?
 
+    private lazy var remoteConfigCacheExpiration: TimeInterval = {
+        if AppConstants.isDebug {
+            // Refresh the config if it hasn't been refreshed in 60 seconds.
+            return 60
```
jhugman: The production builds will only refresh their configs every few hours or so.

The debug builds will likely want their configs refreshed on a edit-compile-run cycle of around 60 seconds.

PR #181, opened 2016-11-16
Closes #159: collapse description views when re-using CardViewVCs.

### LINKED ISSUE #159: Unseen description views already expanded when swiped to
Another ViewController re-use problem, I imagine.

STR:
1) Click Yelp description view
2) Swipe right (maybe 5x?)

Expected: Only the initial card is expanded
Actual: Another card is expanded.

mcomella:


fluffyemily:
Seeing as we're getting a number of these reuse issues we can either:

1. Add a `prepareForReuse` function that we can call as we dequeue a view controller that will then handle all the resetting of the views in one go 
OR
2. Remove the reuse and go with just creating and destroying as we go.

I went with reuse orignally as for large sets of data, scrolling quickly will use a lot of memory as ARC tries to keep up with the rapid creation and destruction of objects and efficiency will drop, so personally I would prefer to keep the reuse if possible, however if you feel it would be simpler just to chuck it, feel free. At this stage I'm not going to get pissy about it.

mcomella:
> I would prefer to keep the reuse if possible,

You know better than I so let's keep it. I'll add a `prepareForReuse` next time I edit this (#196?).

Issue #168, opened 2016-11-14
Travel times loading spinner has forward arrow


mcomella:
A regression from #84, I imagine.

STR:
1) Open app

Expected: Travel times spinner has no arrow
Actual: Has arrow

I don't see it on the simulator but it's pretty apparent on device.

PR #203, opened 2016-11-16
Closes #167: Add error state for travel times.

### LINKED ISSUE #167: Travel times missing (no error state)
Maria noticed this issue:
img_4880

I'd guess the travel times were unable to be fetched but we didn't update the UI correctly. How do we handle the error states?

mcomella:
The forward arrow is not present yet but should come for free when #168 lands.

Post-patch:
simulator screen shot nov 16 2016 15 46 33


fluffyemily:
Needs rebasing before merge.

I also need to figure out what's going wrong with your buddybuild build - looks like the secure file discovery isn't working for forks.

mcomella:
> I also need to figure out what's going wrong with your buddybuild build

I don't think it's worth the time to set up right now – I've been running a pre-push hook to run the tests so I think we'll be fine.

mcomella:
And if we're really concerned, I can work off the main repo.

Issue #193, opened 2016-11-16
Incorporate event search into fetching of places


fluffyemily:
When we fetch places, we need to also fetch events from a much wider radius and augment our place list with the places that contain events

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [Show events happening nearby](https://trello.com/c/Zf5kaT3F/195-show-events-happening-nearby)

PR #205, opened 2016-11-17
Issue 193 fetch events with places

fluffyemily:
#193 

Do a parallel search for events over a wider radius (specified in remote config) than with places. Add the places that contain events to the results from the place search to ensure that the UI has all the places it cares about.

Although we currently key events by place id and therefore can not have more than one event per place, I have also added the ability to handle multiple events for a place as I can imagine this is something that may change.


fluffyemily:
[Show events happening nearby](https://trello.com/c/Zf5kaT3F/195-show-events-happening-nearby)

Prox/Prox/Data/PlacesController.swift
```diff
@@ -40,6 +40,9 @@ class PlacesProvider {
         return FIRRemoteConfig.remoteConfig()[key].numberValue!.doubleValue
     }()
 
+    private var nearbyPlaces = [CLLocation: [Place]]()
```
fluffyemily: We are only caching for the duration of the call. Once we have received a response from both the nearby place search and the places with events search then the cache is cleared. It is there solely to enable us to merge the two sets of places together once we have them.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -110,13 +117,36 @@ class PlacesProvider {
         }
     }
 
-    private func displayPlaces(places: [Place], forLocation location: CLLocation) {
-        let preparedPlaces = self.preparePlaces(places: places, forLocation: location)
+    private func fetchPlacesWithEvents(location: CLLocation) {
+        let eventProvider = EventsProvider()
+        eventsPlaces.removeValue(forKey: location)
```
fluffyemily: I've cleaned this up.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -110,13 +117,36 @@ class PlacesProvider {
...
+        eventProvider.getEventsWithPlaces(forLocation: location, usingPlacesDatabase: database) { places in
+            self.eventsPlaces[location] = places
+            self.displayPlaces(forLocation: location)
+        }
+    }
+
+    private func displayPlaces(forLocation location: CLLocation) {
+        guard let places = self.nearbyPlaces[location],
+            let eventPlaces = self.eventsPlaces[location] else { return }
```
fluffyemily: `displayPlaces` will get called multiple times for `nearbyPlaces` (each time around the loop while we poll to see if the database has been populated yet) and once for `eventPlaces`. The idea had been that we only tell the delegate once that there are new places once we have them all, however I have now updated it so that we tell the delegate every time `displayPlaces` is called, but only merge the results once we have both. Once we have both, we clear the cache and start from fresh.

These two calls currently run concurrently and I would rather that than having them run in sequence seeing as it can take a very long time for the places fetch to return.

I've made this clearer

Prox/Prox/Data/PlacesController.swift
```diff
@@ -110,13 +117,36 @@ class PlacesProvider {
...
+        let preparedPlaces = self.preparePlaces(places: union(ofNearbyPlaces: places, andEventPlaces: eventPlaces), forLocation: location)
```
fluffyemily: events are filtered before they get here. The filtering occurs in the `EventsController`

Issue #194, opened 2016-11-16
Wire up event data with place details cards


fluffyemily:
If a place in Place Details has an event, ensure that it is displayed on the UI. This involves attaching events to places and checking for them when viewing cards

PR #208, opened 2016-11-17
Issue 152 - fire notifications for event

fluffyemily:
This patch and #193 &  #194 are perhaps a little rushed as there is still tons to do and only one day left. I am expecting a _lot_ of comments, but we should at least have _something_ to demo at the end of the sprint if I spend it on the notification criteria parameters.


Prox/Prox/Utilities/EventNotificationsManager.swift
```diff
@@ -6,8 +6,20 @@ import Foundation
 import CoreLocation
 import UserNotifications
 
+private let sentNotificationDictKey = "sent_notifications_dict"
+
 class EventNotificationsManager {
 
+    fileprivate var sentNotifications: [String: [String]] {
+        get {
+            return UserDefaults.standard.dictionary(forKey: sentNotificationDictKey) as? [String : [String]] ?? [String : [String]]()
```
mcomella: We probably don't need to care but I'd expect this to grow infinitely, taking up a lot of disk space.

Perhaps add a TODO so we don't forget?

Prox/Prox/Utilities/EventNotificationsManager.swift
```diff
@@ -6,8 +6,20 @@ import Foundation
 import CoreLocation
 import UserNotifications
 
+private let sentNotificationDictKey = "sent_notifications_dict"
+
 class EventNotificationsManager {
 
+    fileprivate var sentNotifications: [String: [String]] {
```
mcomella: nit: say this is placeID to event description

Prox/Prox/Utilities/EventNotificationsManager.swift
```diff
@@ -134,4 +149,21 @@ class EventNotificationsManager {
...
+        }
+        return !placeEvents.contains(event.description)
+    }
+
+    fileprivate func markAsSent(event: Event) {
+        var sent = sentNotifications
+        if sent.keys.contains(event.placeId) {
+            sent[event.placeId]?.append(event.description)
+        } else {
+            sent[event.placeId] = [event.description]
```
mcomella: nit: Not sure we can store in UserDefaults but this would be more efficient as a set (at least, as these grow larger and hashing time is more efficient than array search).

mcomella: nit: I'm not sure `event.description` is guaranteed not to change – maybe `event.url` is slightly safer? In any case, in the future, it'd be good to create event IDs for our internal use.

Prox/Prox/Data/EventsController.swift
```diff
@@ -52,8 +52,9 @@ class EventsProvider {
 
     private func doesEvent(event: Event, startAtCorrectTimeIntervalFromNow timeInterval: TimeInterval) -> Bool {
```
fluffyemily: correct is the time that is passed in. Will rename.

Prox/Prox/Utilities/EventNotificationsManager.swift
fluffyemily: Good point. I'm sure we can use the start time of the event at a place to resolve this, and clear out anything from the cache that is not today. I'll also use an NSCache and only store to `UserDefaults` when we go into the background, rather than storing in `UserDefaults` by default

Issue #157, opened 2016-11-10
Score missing from TA on map view


mcomella:
Looks like a regression from #141, a8b12e81974820bcca666b23e9a755669f144aff – I checked out the parent commit and the issue no longer appears.

![simulator screen shot nov 10 2016 11 01 51](https://cloud.githubusercontent.com/assets/759372/20190184/4495fb2e-a735-11e6-8e85-4fab4b0d6ce6.png)

mcomella:
I think the UX #141 provides is not what was designed – I [asked Ricardo in the error state bug](https://github.com/mozilla-mobile/prox/issues/38#issuecomment-260810788) what is desired. 

In the best case, we can just backout #141, which should fix this bug and maybe #38. That being said, also verify *view re-use is happening correctly* – the missing score video shouldn't be happening and this is a likely cause.

mcomella:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user, I want the places view to handle missing either yelp reviews or yelp ratings (we already handle missing both)](https://trello.com/c/4Xn9BZ37/205-as-a-user-i-want-the-places-view-to-handle-missing-either-yelp-reviews-or-yelp-ratings-we-already-handle-missing-both)

PR #211, opened 2016-11-18
#38, #157: Fix review provider error states

mcomella:
#38, #157


mcomella:
[As a user, I want the places view to handle missing either yelp reviews or yelp ratings (we already handle missing both)](https://trello.com/c/4Xn9BZ37/205-as-a-user-i-want-the-places-view-to-handle-missing-either-yelp-reviews-or-yelp-ratings-we-already-handle-missing-both)

PR #214, opened 2016-11-18
Issue 153 respond to notification

fluffyemily:
This is, again, rough around the edges. I imagine a fair amount of next sprint will be spent iterating on it and getting it right.

I've got this on my phone so will spend some time this weekend seeing what happens and how it reacts.

This really needs some tests in the right places, but left out for time purposes. 

I know!

Stop looking at me that way!


Prox/Prox/Data/EventsController.swift
```diff
@@ -43,7 +47,8 @@ class EventsProvider {
     }
 
     private func shouldShowEventForNotifications(event: Event, forLocation location: CLLocation) -> Bool {
-        return doesEvent(event: event, startAtCorrectTimeIntervalFromNow: eventStartNotificationInterval)
+        return isEventToday(event: event) && isEventYetToHappen(event: event)
+//        return doesEvent(event: event, startAtCorrectTimeIntervalFromNow: eventStartNotificationInterval)
```
mcomella: nit: rm

Prox/Prox/Data/PlacesController.swift
```diff
@@ -47,6 +47,29 @@ class PlacesProvider {
         database.getPlace(forKey: key).upon { callback($0.successResult() )}
     }
 
+    func placeWithEvent(forKey key: String, callback: @escaping (Place?) -> ()) {
```
mcomella: for future use: perhaps rather than accessing the DB each time we need to get a place or an event, since we already have the places/events in memory to show in the UI, we could get them from in memory.

Would prevent issues where the data we get from the DB is newer than what's in memory and we try to match that new data to the old data in memory.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -47,6 +47,29 @@ class PlacesProvider {
...
+        let eventProvider = EventsProvider()
+        var placeWithEvent: Place?
+        var eventForPlace: Event?
+        place(forKey: key) { place in
+            guard let foundPlace = place,
+                let event = eventForPlace else {
+                return placeWithEvent = place
```
mcomella: nit: we don't return a value so it's confusing to have these on the same line

Prox/Prox/Data/PlacesController.swift
```diff
@@ -47,6 +47,29 @@ class PlacesProvider {
...
+            }
+            foundPlace.events.append(event)
+            callback(foundPlace)
+        }
+
+        eventProvider.event(forKey: key) { event in
```
mcomella: Can these run concurrently? If so, we need locking.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -191,6 +192,7 @@ class PlaceCarouselViewController: UIViewController {
     }
 
     override func viewWillAppear(_ animated: Bool) {
+        print("viewWillAppear")
```
mcomella: nit: want to keep these? There's one for viewDidAppera too

Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -242,6 +242,83 @@ class PlaceDetailViewController: UIViewController {
         // Dispose of any resources that can be recreated.
     }
 
+    func openCard(forPlace place: Place) {
+        if currentCardViewController.place == place {
+            currentCardViewController.place = place
+            // if we're currently looking at that card then just animate in the event
+            currentCardViewController.showEvent()
+        } else {
+            // I'm totally not sure what we should be doing here.
```
mcomella: nit: how should we track this? (if not fixed in a later commit).

Is it that issue Ricardo is working on? Maybe list that here.

Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -242,6 +242,83 @@ class PlaceDetailViewController: UIViewController {
         // Dispose of any resources that can be recreated.
     }
 
+    func openCard(forPlace place: Place) {
+        if currentCardViewController.place == place {
+            currentCardViewController.place = place
```
mcomella: nit: redundant.

Is this to refresh the view?

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -271,6 +271,12 @@ class PlaceDetailsCardView: UIView {
         }
     }
 
+    func showEvent(atPlace place: Place) {
+        updateEventUI(forPlace: place)
+        self.setNeedsLayout()
+        self.layoutIfNeeded()
```
mcomella: I think these should be inside updateEventUI.

In that case, why not just expose `updateEventUI` and get rid of the callthrough?

Prox/Prox/Utilities/EventNotificationsManager.swift
```diff
@@ -151,10 +153,10 @@ class EventNotificationsManager {
     }
 
     fileprivate func isUnsent(event: Event) -> Bool {
-        guard let placeEvents = sentNotifications[event.placeId] else {
+        // guard let placeEvents = sentNotifications[event.placeId] else {
```
mcomella: Test code?

Prox/Prox/Utilities/LocationMonitor.swift
```diff
@@ -141,7 +141,10 @@ class LocationMonitor: NSObject {
 
 extension LocationMonitor: LocationProvider {
     func getCurrentLocation() -> CLLocation? {
-        return currentLocation ?? locationManager.location
+        if currentLocation == nil {
+            currentLocation = self.locationManager.location
```
mcomella: iirc, this location may not be have been processed by the locationManager delegate yet – are we okay with that? Are there any assumptions in the code that could cause problems?

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -23,6 +23,9 @@ struct PlaceUtilities {
 
     static func filterPlacesForCarousel(_ places: [Place]) -> [Place] {
         return places.filter { place in
+            // always show places if they have events
+            guard place.events.isEmpty else { return true }
```
mcomella: Are these events guaranteed to be soon enough that we want to show them? Or is this just for testing? If just testing, it could be good to add a comment to that effect.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: No, left over debugging caused by rushing and tiredness

Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
fluffyemily: No redundant. The place that the card had before this function was called had no event. The place triggering this function has an event, therefore is necessary. I'll rename things to make that clearer

Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
fluffyemily: left over debug. The question wasn't from a UI perspective, but from an architectural one. Will remove

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
fluffyemily: They were here because I am going to add an animation to make the event appear nicer, and I don't really want the animation in the event view as it will need to be in the card view. I shall add comments to that effect here

Prox/Prox/Utilities/LocationMonitor.swift
fluffyemily: Yup, the rest of the code handles the nil location case, this is just resetting the currentLocation on start

Prox/Prox/Utilities/PlaceUtilities.swift
fluffyemily: Yes, only events that are to be shown will appear on places. If an event shouldn't be shown, it will no appear on a place as they are filtered out in the `EventProvider`

Issue #175, opened 2016-11-15
Loading Screen IF wait is longer than 60 seconds


vazquez:
Upon first launch of the app, we show a loading screen while we populate the information in the background. In the event that we can't populate event data, we need to show a message inviting people to restart this Beta App.

![home - tropics ale house copy 7](https://cloud.githubusercontent.com/assets/1119821/20352417/33d39c4c-abe5-11e6-94e9-e23df8184de8.png)

![home - tropics ale house copy 7](https://cloud.githubusercontent.com/assets/1119821/20352603/d6977f70-abe5-11e6-9620-bc01c97171ee.png)

fluffyemily:
![](https://github.trello.services/images/mini-trello-icon.png) [Loading screen improvements](https://trello.com/c/fed9HUKu/180-loading-screen-improvements)

sleroux:
@vazquez What do you think about fading in the label/buttons like dis

![error_fade_in](https://cloud.githubusercontent.com/assets/434322/20409458/ab27d314-ace7-11e6-8eb6-c746a6b2bcad.gif)

vazquez:
Yeah Steph! Let's do that 🙂

Issue #131, opened 2016-11-08
Loading Screen Missing


vazquez:
It seems the loading screen is missing from the current Prox build.

This is the loading animation we have explored in Framer: http://share.framerjs.com/9n2t6pb3umqz/

![screenshot 2016-11-08 15 18 19](https://cloud.githubusercontent.com/assets/1119821/20115535/9c261dda-a5c6-11e6-834e-970aaf5458bf.png)

fluffyemily:
I was unaware of any loading screen or animation being provided to us yet.
Please add specs

vazquez:
## Interaction

GIF showing interaction:

![loading animation](https://cloud.githubusercontent.com/assets/1119821/20148865/8e652d60-a67c-11e6-903d-c876f61347be.gif)

## Assets

- Background – `background_loading.png`
- Blue Loading Dot – `icon_loadingdot.png`

The animation works like this:

- Two circles are positioned at the center
- The sizes of both circles are `height: 100px` and `width: 100px`
- CircleA and CircleB have a background colour of `#0096DD`
- CircleA and CircleB start at `opacity: 0.12`
- CircleA grows to `120px` and changes its opacity to `0.1` while circleB shrinks to `90px` and changes it opacity to `0.3`
- The animation loops back and forth.

mcomella:
![](https://github.trello.services/images/mini-trello-icon.png) [Loading screen improvements](https://trello.com/c/fed9HUKu/180-prox-starts-like-in-this-screenshot-and-stays-there-for-a-really-long-time-like-30-seconds-after-that-it-switches-to-the-card-vi)

Issue #218, opened 2016-11-19
`PlacesController.updatePlaces` is called frequently, with no short-circuit


mcomella:
`PlacesController.`updatePlaces is called every time `PlaceCarouselViewController.viewWillAppear` is called.

Each time `updatePlaces` is called, we ping the server with our location, query the firebase, and replace all our in-memory places with a new set. In my particular case, I was trying to add a travel times cache on the places and they'd get wiped out each time, requiring recomputation.

Since we don't have firebase persistence enabled, it's also possible each time we query the firebase, we'll pulling down redundant data.

We should also check to see if there's anywhere else the places may be getting replaced from (e.g. place details, events).

CC @fluffyemily for awareness.

fluffyemily:
update places will only get called when the app opens and when you close the details view. We should add a check to ensure that the location has changed significantly since the last fetch to ensure we are not fetching for a location we already have data for.

mcomella:
I haven't dug in but it seems that the set of places gets updated twice just from launching the app.

jhugman:
>
> Since we don't have firebase persistence enabled, it's also possible each
> time we query the firebase, we'll pulling down redundant data.

This is a worry. I was under the impression that persistence was enabled.

I haven't dug in but it seems that the set of places gets updated twice
> just from launching the app.
>
This is once at `viewDidAppear` and once when the location is set. We
prevent a race by only allowing one update to happen at a time.

fluffyemily:
I am resolving this _right now_

mcomella:
@fluffyemily Did you close this because it was fixed in #226?

fluffyemily:
Yes

PR #226, opened 2016-11-21
Issue 131/175 - Animated loading screen w/ error state

sleroux:
https://github.com/mozilla-mobile/prox/issues/131
https://github.com/mozilla-mobile/prox/issues/175

* Adds location loading animation to a loading overlay view
* Loading overlay view can animate in the error text/button when user timeouts

The timing out functionality isn't implemented yet and the loading screen appears below the 'Waiting'/sunview headers. Not sure what we want to do with those while we're waiting.


fluffyemily:
[Loading screen improvements](https://trello.com/c/fed9HUKu/180-loading-screen-improvements)

sleroux:
I probably should have marked this as incomplete. 

> Either add a blur background to the LoadingOverlayView and present it above the header and sun views OR hide the header and sun views and show them only as we dismiss the LoadingOverlayView

I think hiding the header/sunview would be nicer. When those finish loading along with the places we can cross fade the loading overlay into the header + places.

> Remove the 'WaitingandNo Places Found` text that we added as a temporary stopgap until this ticket was implemented
👍

I forgot to ask but currently we have the button 'Restart Prox'. What does restart mean in this case? I don't think we want to force close the app in any case. I'm not sure that this applies to mobile. Is there some other behavior that should be happening that I'm missing?

fluffyemily:
> I forgot to ask but currently we have the button 'Restart Prox'. What does restart mean in this case? I don't think we want to force close the app in any case. I'm not sure that this applies to mobile. Is there some other behavior that should be happening that I'm missing?

@vazquez what was your intention with the 'Restart Prox' button functionality?

sleroux:
Somewhat related but I figured I'd ask here anyways:

Is having the closest place open automatically the intended behavior? It feels a little too 'automatic' for me to have the app just open what's close without giving me a chance to scan what's nearby. I feel like there's more value in what's around me than what's the closest option.

fluffyemily:
@sleroux The idea of the app is that we always show you the most useful place first. The way we determine the most useful place may change, but right now it's defined as the closest place, which will often be the place you are currently at.

This functionality is the core principle behind the app, so it's super interesting that you don't like it. I think this is worth sticking in the slack channel as a conversation starter for future thinking.

vazquez:
@sleroux @fluffyemily –

What I gathered from my chat with @liuche was that upon first-launch of the app, we query all of our services. If we get back nothing, then we try again. However, we have a limited amount of tries (3?), and once we exhaust all tries the app is stuck in limbo, in a perpetual loading screen.

Rather than adding a message saying "Please restart Prox" (which would clear everything and start over), I added a button instead. This functionality is a way for us to get around the limitations of calling services a limited amount of times.

@liuche Correct me if I'm wrong!

fluffyemily:
@vazquez - but we cannot restart Prox for the user. We can force close it,
which is not a good thing to do and, if we were distributing through the
app store, may result us being rejected on review, but we cannot restart it
again.

Hence the question, what do you want to actually happen when the user
presses the "restart prox" button?

vazquez:
I want our service query limitations to be reset back to zero so we can query our API's again.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -41,11 +41,6 @@ class PlaceCarouselViewController: UIViewController {
             // TODO: how do we make sure the user wasn't interacting?
             headerView.numberOfPlacesLabel.text = "\(places.count) place" + (places.count != 1 ? "s" : "")
             placeCarousel.refresh()
-
-            if oldValue.isEmpty,
```
sleroux: I was experiencing some weird timing/navigation issues on my transition animation branch with this here so I moved it out of the setter and call it explicitly after we set places in `didReceivedPlaces` below

sleroux:
Some comments about Part 3/4:

I haven't hooked up the button yet to 'restart' Prox so the button doesn't do anything. I also fiddled a bit with how and when we load data. This could not make sense so let me know if this doesn't work but the flow I was going after was:

1. On first load (viewDidLoad), get the current location, throw up the loading UI, and fetch places
2. Whenever the user returns back to the carousel screen (viewWillAppear), don't explicitly update places but see if we should start monitoring our location. If we start monitoring and the location has been updated, places will be updated through `didUpdateLocation`. 

Essentially this means that the data we present will be cached until the user moves to ~50m (I think is what the radius is set to) or until the app is restarted. This might not be ideal considering that I can see local data around me updating somewhat-frequently so maybe we can schedule a refresh timer instead of having the cache lifecycle correspond with the `viewWillAppear`? 

To keep the data consistent between the places on the map and the places in the detail carousel, we could move to using a `NSNotification` for data updates and broadcast to both at an interval. This would help eliminate some of the flickering/views invalidating every time the user navigates back to the map view.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -189,22 +196,20 @@ class PlaceCarouselViewController: UIViewController {
...
+            updatePlaces(forLocation: location)
+        }
     }
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
-        guard let location = locationMonitor.getCurrentLocation() else {
-            return
+        if shouldStartMonitoringLocation() {
```
fluffyemily: #218 is addressing the place fetching on location update issues. I don't want competing implementations causing problems on merge

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -10,7 +10,7 @@ import Deferred
 
 private let MAP_SPAN_DELTA = 0.05
 private let MAP_LATITUDE_OFFSET = 0.015
-
+private let LOADING_TIMEOUT_IN_SECONDS: TimeInterval = 10
```
fluffyemily: Place Loading times out when `placesProviderDidFinishFetchingPlaces` is called, and the amount of time it continues to try and fetch places for is specified in `PlacesProvider.numberOfRetries` (which really should be moved to remote config). It is currently set to 60, which means the place fetching will try every second for `numberOfRetries` for as long as we are receiving either no places or the number of places we are fetching continues to rise. This is so that if it is taking the server a long time to gather the data for an area and populate the database we continue to try while we fetch.

With this timeout implementation we could tell people we found no places after 10 seconds and then start receiving data after 11 seconds. I'm not sure how that would feel.

It makes me wonder whether the place fetching timeout should be moved to `PlacesProvider`. If we have retried 10 times without receiving any places, stop fetching places and call `placesProviderDidFinishFetchingPlaces`. If we have started to receive places we can continue to loop in our fetching until we have received them all.

Does that make sense?

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -368,27 +387,27 @@ extension PlaceCarouselViewController: EventsProviderDelegate {
...
-        // TODO placeholder for the waiting state.
-        if self.places.count == 0 {
-            headerView.numberOfPlacesLabel.text = "Waiting"
-        }
     }
 
     func placesProviderDidFinishFetchingPlaces(_ controller: PlacesProvider) {
-        // no op
         if self.places.isEmpty {
             headerView.numberOfPlacesLabel.text = "No Places Found"
```
fluffyemily: should we also be calling `fadeInMessaging` here?

sleroux: I guess it depends if we want to use the loading error state as the empty state. I can see this being shown if we haven't received any places - like on first load - but might be strange if we are replacing cached sites with this.

sleroux:
@fluffyemily Also just followingup on @vazquez's answer

> I want our service query limitations to be reset back to zero so we can query our API's again.

Is this simply restarting our request to PlacesProvider or is there something else that I'm missing?

fluffyemily:
> >  I want our service query limitations to be reset back to zero so we can query our API's again.
>Is this simply restarting our request to PlacesProvider or is there something else that I'm missing?

I think so. We might want to rethink the text though as restarting Prox isn't what we're doing. Maybe something along the lines of "Search again"?

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: But if we've gone from having places to having no places, I think we need to tell the user that there are no places in their new location rather than just leaving them with a blank app.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: >   I want our service query limitations to be reset back to zero so we can query our API's again.

Is this simply restarting our request to PlacesProvider or is there something else that I'm missing?

Hmm, I did  assume that we would show this every time we started loading places, but maybe that is a question for @vazquez

vazquez:
Rewording the text to "Search again" would be awesome :+1:

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: > I agree this makes the most sense. Maybe add a new delegate method named placesProviderDidTimeout to make it more explicit and add the timeout UI update there?

Sounds like a plan

vazquez:
home - tropics ale house copy 9

Prox/Prox/Data/PlacesController.swift
```diff
@@ -106,6 +107,7 @@ class PlacesProvider {
             failure: { (task, err) in
                 NSLog("Error from server: \(err)")
                 DispatchQueue.main.async {
+                    self.isUpdating = false
```
sleroux: When I was testing I saw that we weren't resetting the `isUpdating` flag on errors. Added a fix for that.

sleroux:
@fluffyemily Should be good for another go

Prox/Prox/Data/PlacesController.swift
```diff
@@ -126,8 +128,15 @@ class PlacesProvider {
         database.getPlaces(forLocation: location, withRadius: radius).upon { results in
             let places = results.flatMap { $0.successResult() }
             let placeCount = places.count
+
+            guard retriesLeft > 0 else {
+                self.isUpdating = false
+                self.delegate?.placesProviderDidTimeout(self)
```
fluffyemily: Either this needs to be called from the main thread, or the delegate needs to perform it's actions on the main thread once it's called as I'm getting a load of these warnings:

```
2016-11-25 09:54:52.235 Prox[31633:645351] This application is modifying the autolayout engine from a background thread after the engine was accessed from the main thread. This can lead to engine corruption and weird crashes.
 Stack:(
	0   CoreFoundation                      0x000000010a6f634b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x000000010a15721e objc_exception_throw + 48
	2   CoreFoundation                      0x000000010a75f265 +[NSException raise:format:] + 197
	3   Foundation                          0x0000000109e522b8 _AssertAutolayoutOnAllowedThreadsOnly + 180
	4   Foundation                          0x0000000109e52000 -[NSISEngine _optimizeWithoutRebuilding] + 61
	5   Foundation                          0x0000000109c79dfc -[NSISEngine optimize] + 91
	6   Foundation                          0x0000000109e4ff20 -[NSISEngine performPendingChangeNotifications] + 84
	7   UIKit                               0x000000010b979b16 -[UIView(Hierarchy) layoutBelowIfNeeded] + 265
	8   Prox                                0x0000000107e3f3a5 _TFC4Prox18LoadingOverlayView15fadeInMessagingfT_T_ + 53
	9   Prox                                0x0000000107e288e1 _TFC4Prox27PlaceCarouselViewControllerP33_E0A168B46C9580E142C881F80CE2EF0926showErrorMessageIfNoPlacesfT_T_ + 145
	10  Prox                                0x0000000107e2900c _TFC4Prox27PlaceCarouselViewController24placesProviderDidTimeoutfCS_14PlacesProviderT_ + 28
	11  Prox                                0x0000000107e29171 _TTWC4Prox27PlaceCarouselViewControllerS_22PlacesProviderDelegateS_FS1_24placesProviderDidTimeoutfCS_14PlacesProviderT_ + 49
	12  Prox                                0x0000000107e468a0 _TFFC4Prox14PlacesProviderP33_34802027475AD4230BD0432332D06C1516retryQueryPlacesFT8locationCSo10CLLocation10withRadiusSd11retriesLeftSi9lastCountSi_T_U_FGSaGCS_14DatabaseResultCS_5Place__T_ + 2096
	13  Prox                                0x0000000107dc8638 _TTRXFo_oGSaGC4Prox14DatabaseResultCS_5Place____XFo_iGSaGS0_S1_____ + 24
	14  Deferred                            0x00000001098e1735 _TFFC8Deferred8DeferredP33_8471EEB727F82F3A241B76FBEA842E6D6notifyFT5flagsV8Dispatch21DispatchWorkItemFlags4uponCSo13DispatchQueue7executeFxT__T_U_FT_T_ + 469
	15  Deferred                            0x00000001098e1b27 _TTRgrXFo___XFdCb___ + 39
	16  libdispatch.dylib                   0x000000010db2b278 _dispatch_block_async_invoke_and_release + 75
	17  libdispatch.dylib                   0x000000010db500cd _dispatch_client_callout + 8
	18  libdispatch.dylib                   0x000000010db31691 _dispatch_root_queue_drain + 1450
	19  libdispatch.dylib                   0x000000010db3108b _dispatch_worker_thread3 + 123
	20  libsystem_pthread.dylib             0x000000010def94de _pthread_wqthread + 1129
	21  libsystem_pthread.dylib             0x000000010def7341 start_wqthread + 13
```

fluffyemily:
```
DispatchQueue.main.async {
    self.delegate?.placesProviderDidTimeout(self)
}
```

Issue #178, opened 2016-11-15
Finalize Yelp categories to restrict


mcomella:
Follow-up to #49. Once it lands, we should see what UX thinks and see where else we can edit the category filters.

After #49, we hide a place if its categories match *all* of the following categories:

> "auto",
                                                  "bicycles",
                                                  "education",
                                                  "financialservices",
                                                  "health",
                                                  "homeservices",
                                                  "localservices",
                                                  "professional",
                                                  "realestate",
                                                  "religiousorgs")

In a 4km radius from the hotel, 0 places were filtered under this criteria.

mcomella:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user I want to see the most useful places when I open the app so that I can make better decisions about my next step](https://trello.com/c/IBEHIXNb/149-as-a-user-i-want-to-see-the-most-useful-places-when-i-open-the-app-so-that-i-can-make-better-decisions-about-my-next-step)

mcomella:
Next steps:
* UX will go through the full category list (yelp API v2 or v3 TBD) to identify which sub-categories we might want to block
* UX will decide if we to hide places if one of their categories matches a UX-specified hide category or if *all* of their categories match UX-specified hide categories
* With list ^, eng to implement

Engn notes:
* We're whitelisting the top-level categories
  * We can filter the initial yelp query by these categories

mcomella:
Note to self: remoteConfig! If you implement this soon, UX can update everything themselves (assuming the logic doesn't need to change).

mcomella:
Spoke with @vazquez:
* UX will continue to create a list of categories we want to hide.
  * If UX specifies a category, I will hide that category *and all of its children*. This is for simplicity: UX has to specify fewer things and we don't need more complex logic (e.g. "Hide category except for children...")
  * I will add a parameter to Firebase remote config where we can dump a csv of categories and the client will hide that list of categories. This will allow the designers to quickly update the category list themselves.
    * I gave Ricardo [a script to convert a newline-separated file into a csv](https://gist.github.com/mcomella/142b8339e34b24a02d6233bc3cb93a42) to make creating a csv easier
* Still TBD: UX will decide if we to hide places if one of their categories matches a UX-specified hide category or if all of their categories match UX-specified hide categories

Yelp API investigations:
* If a top-level Yelp category has children, places can still have that top-level category as a category.

mcomella:
[...]

PR #239, opened 2016-11-22
Issue #178: Take CSV of categories to hide from Remote Config

mcomella:
Yay, design can modify the categories themselves!

After this, #178 will become a UX-only bug (assuming they don't need anymore eng intervention).


Prox/Prox/Utilities/CategoriesUtil.swift
```diff
@@ -6,51 +6,81 @@ import Foundation
...
+        let allMatched = categorySet.subtracting(HiddenCategories).isEmpty
+        return !allMatched
+    }
+
+    // UX gives us a CSV list of categories to hide and we hide them *and their children*.
+    // This is an exhaustive collection of these categories and their children.
+    // Caveat: we don't expect UX to include root categories, which are special-cased: we can reduce
+    // yelp resource use by whitelisting the categories we ask for meaning the other root categories
+    // are hidden implicitly.
+    static let HiddenCategories: Set<String> = {
```
fluffyemily: nit: `hiddenCategories`. I know I've been bad at enforcing this but it's a new Swift 3.0 thing that constants are lowerCamelCase and it's taken me a while to get used to it.

Prox/Prox/Utilities/CategoriesUtil.swift
```diff
@@ -13,6 +13,36 @@ struct CategoriesUtil {
...
+        let csv = RemoteConfigKeys.getString(forKey: RemoteConfigKeys.placeCategoriesToHideCSV)
+        return getHiddenCategories(forCSV: csv)
+    }()
+
+    // Separated for testing: I don't know how to do automated tests with the Firebase value.
+    internal static func getHiddenCategories(forCSV csv: String) -> Set<String> {
+        // We could be more robust
+        let categories = csv.characters.split(separator: ",")
+
```
jhugman: In #241 I have added a method get string arrays out of RemoteConfig. This helps with ensuring accessing and splitting the string, and insist on setting a default.

Prox/Prox/RemoteConfigDefaults.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>place_categories_to_hide_csv</key>
+	<string></string>
```
mcomella: For now, yes – we don't know for sure what categories we want to filter so we should start at nothing.

mcomella:
Waiting for #241 to land before landing this.

Prox/Prox/RemoteConfigDefaults.plist
mcomella: > ...but on second thought, maybe I should also filter on the client so we can get accurate data without relying on the server being correct.

This is less important than other things I can do so filed follow-up #251.

Prox/Prox/Utilities/CategoriesUtil.swift
mcomella: Decided not to use it because it'd be more work to switch over the already working code.

mcomella:
This is blocked on #261. I don't want to rebase and land on top because it'll make backing it all out more difficult.

PR #241, opened 2016-11-22
Closes #228 – Harden RemoteConfig, to make resiliant to configuration error related catastrophe.

### LINKED ISSUE #228: Harden Remote Config access to reduce catastrophic failure
Right now, if an admin fat fingers or adds a value that cannot be parsed by Swift, then there is a explicit dereference of values which will cause a crash. 

e.g. https://github.com/mozilla-mobile/prox/blob/master/Prox/Prox/Data/PlacesController.swift#L40

```swift
```

Worse, the crash may occur before a corrected (non-crashing) config can be downloaded and applied, so the app will be unlaunchable.

Each use of remote config must be guarded properly. 

Proposal is to wrap all access of RemoteConfig, and the defaults into a property class, in `RemoteConfigKeys`

```swift
let radius: RemoteConfigProperty<Double> = RemoteConfigDouble("radius", 40) 
```

And used: 
```swift
let radius = RemoteConfigKeys.radius.value
```

jhugman:
This changes the process of specifying and using values: 

 * defaults, where possible should be specified close to the key.
 * a type is specified at time of declaration.
 * use of the defaults plist is deprecated, though support is not removed, since we are constructing keys in code.

This also adds support for String arrays, which will be useful in categories.


Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -28,36 +28,40 @@ class RemoteConfigKeys {
     // Any previously fetched and cached config would be considered expired because it would have been fetched
     // more than remoteConfigCacheExpiration seconds ago. Thus the next fetch would go to the server unless
     // throttling is in progress. The default expiration duration is 43200 (12 hours).
-    public static let remoteConfigCacheExpiration = "remote_config_cache_expiration"
+    public static let remoteConfigCacheExpiration = RemoteConfigDouble(key: "remote_config_cache_expiration", defaultValue: 12.0 * 60 * 60)
```
mcomella: nit: feel like we should use some constants like `TIME_IN_HOUR` here or something: multiplying by values of time always feels so fragile!

Prox/Prox/Data/EventsController.swift
```diff
@@ -15,15 +15,15 @@ class EventsProvider {
     lazy var eventsDatabase: EventsDatabase = FirebaseEventsDatabase()
 
     private lazy var radius: Double = {
-        return RemoteConfigKeys.getDouble(forKey: RemoteConfigKeys.eventSearchRadiusInKm)
+        return RemoteConfigKeys.eventSearchRadiusInKm.value
     }()
 
     private lazy var eventStartNotificationInterval: TimeInterval = {
-        return RemoteConfigKeys.getTimeInterval(forKey: RemoteConfigKeys.eventStartNotificationInterval)
+        return RemoteConfigKeys.eventStartNotificationInterval.value * 60
```
mcomella: Maybe we should have `RemoteConfigTimeInterval` or `RemoteConfigMinutesToSeconds` that'd do this fragile multiplication for us?

fluffyemily: I agree.

jhugman: I considered this, but separating how the constant is stored, and how its used seemed to be a useful distinction here. 

Suggest an `TimeInterval` `extension` to manage converting to seconds from minutes. e.g. 

```swift
extension TimeInterval {
    init(minutes: Double) {
        self.init(minutes * 60)
    }

    init(hours: Double) {
        self.init(minutes: hours * 60)
    }
}

```

Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -70,19 +71,62 @@ class RemoteConfigKeys {
...
+        let remoteConfig = FIRRemoteConfig.remoteConfig()
+        let rcv = remoteConfig[key]
+        if let value = convert(rcv) {
+            return value
+        }
+        if let rcv = remoteConfig.defaultValue(forKey: key, namespace: nil),
+            let value = convert(rcv) {
+            return value
+        }
+        return defaultValue
```
jhugman: One comes from swift, one comes from the plist.

I'm not altogether happy with having a plist at all, but current usage for an array of very long strings is done my constructing keys. Rather than a larger scale refactor of that, I commented on this, and continued with supporting the plist.

I've added a print statement to show where the default is coming from.

Issue #198, opened 2016-11-16
Handle updating UI for place pop-in (e.g. map view places count)


kbrosnan:
While in the card view the button for the map view showed an incorrect number of results. Likely hard to reproduce with a cached location.

* First run on a new location 
* Clicked on early results that appeared in the default view
* More results popped in and I was able to swipe between them in the card view
* The map view button stayed at 2 when there were obviously more than 2 items listed

* going back to the map view updated the count to the correct number

mcomella:
> More results popped in and I was able to swipe between them in the card view

We originally wrote the code to expect static places data. However, pop-in is an issue now that we do custom location querying – we have to make sure we handle updating all aspects of the UI, not just place count.

PR #244, opened 2016-11-22
Issue 218 - place update fixes

fluffyemily:
New place update strategy

- When app loads, take current location and monitor a region with a radius of search_radius_in_km/4 kms around that location. 
- Set Flag saying don't fetch places
- When user exists that monitored region, set flag saying fetch places
- If the user is in the background, dismiss the detail view, triggering a fetch of new places immediately
- When user next returns to the carousel screen, fetch new places
- If user opens app or returns to the carousel screen and no new places need to be fetched, resort the current places based on the users location

My simulator has stopped notifying the app on location changes so I have not been able to test this. Will spend tomorrow riding around London on a bus ensuring that this works as expected, but I thought I would raise the PR so the code can get reviewed while I am doing that.

Also tagged both @jhugman & @mcomella to get more eyes on the code

### DIFF
--- a/Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
+++ b/Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
@@ -243,7 +243,7 @@ class PlaceCarouselViewController: UIViewController {
 
     fileprivate func fetchPlaces(forLocation location: CLLocation) {
         self.placesProvider.updatePlaces(forLocation: location)
-        let placeMonitoringRadius = RemoteConfigKeys.getDouble(forKey: RemoteConfigKeys.searchRadiusInKm) / 4
+        let placeMonitoringRadius = RemoteConfigKeys.searchRadiusInKm.value / 4
         locationMonitor.startMonitoring(location: location, withIdentifier: placesFetchMonitorIdentifier, withRadius: placeMonitoringRadius, forEntry: nil, forExit: { region in
             self.locationMonitor.stopMonitoringRegion(withIdentifier: placesFetchMonitorIdentifier)
             self.shouldFetchPlaces = true
--- a/Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
+++ b/Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
@@ -91,7 +91,7 @@ class PlaceCarouselViewController: UIViewController {
         return carousel
     }()
 
-    fileprivate var shouldUpdatePlaces: Bool = true
+    fileprivate var shouldFetchPlaces: Bool = true
 
     var sunriseSet: EDSunriseSet? {
         didSet {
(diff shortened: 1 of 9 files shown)


mcomella:
> If user opens app or returns to the carousel screen

Why do we need to fetch new places when the user enters the carousel screen? Could we just fetch when the user has moved outside of the given region? If that's too infrequent (or not reliable), we could also set a timeout (e.g. when user hasn't used app in an hour).

Refreshing everytime we show the carousel seems more complicated (e.g. we have to deal with pop-in regularly, user can open details and carousel multiple times in a single session, making many requests)

Prox/Prox/RemoteConfigDefaults.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>significant_location_change_distance_meters</key>
+	<integer>100</integer>
```
mcomella: This will need to be rebased on James' changes where the default value is specified in code.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -248,6 +251,22 @@ class PlaceCarouselViewController: UIViewController {
 
     fileprivate func updatePlaces(forLocation location: CLLocation) {
         self.placesProvider.updatePlaces(forLocation: location)
+        let placeMonitoringRadius = RemoteConfigKeys.getDouble(forKey: RemoteConfigKeys.searchRadiusInKm) / 2
```
mcomella: Rather than `searchRadius / 2`, should we just make the place monitoring radius its own remote config value?

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -248,6 +251,22 @@ class PlaceCarouselViewController: UIViewController {
...
+            self.locationMonitor.stopMonitoringRegion(withIdentifier: placesFetchMonitorIdentifier)
+            self.shouldUpdatePlaces = true
+            // if we're currently in the background then dismiss the detail view if we have one
+            // this is so that we get fresh places when we open the app again
+            // right now I have no idea if this will kick of a places fetch immediately or when the app is next opened
+            // maybe I'll have to make some changes once I have figured that out
+            let state = UIApplication.shared.applicationState
+            if state == .background || state == .inactive {
+                DispatchQueue.main.async {
+                    self.presentedViewController?.dismiss(animated: false, completion: nil)
```
mcomella: Is UX okay with the app opening in the app carousel state?

fluffyemily:
This also now addresses #198 as well

PR #256, opened 2016-11-23
Add Flurry analytics

liuche:
Opening a pull request for feedback.

After adding the newly required plist, Prox will send events to Flurry.

Still to do:
* Add unique card view count for Details or Carousel
* Investigate why some timed event durations are 0 in the Flurry dashboard
* Add remote config load timer?

### DIFF
--- a/Prox/Prox/PlaceDetails/PlaceDetailsCardViewController.swift
+++ b/Prox/Prox/PlaceDetails/PlaceDetailsCardViewController.swift
@@ -148,6 +148,8 @@ class PlaceDetailsCardViewController: UIViewController {
            let url = URL(string: urlString) else { return }
         if !OpenInHelper.open(url: url) {
             print("lol unable to open web address")
+        } else {
+            Analytics.logEvent(event: AnalyticsEvent.WEBSITE, params: [:])
         }
     }
 
@@ -155,6 +157,8 @@ class PlaceDetailsCardViewController: UIViewController {
         guard let url = URL(string: place.yelpProvider.url) else { return }
         if !OpenInHelper.open(url: url) {
             print("lol unable to open yelp review")
+        } else {
+            Analytics.logEvent(event: AnalyticsEvent.YELP, params: [:])
         }
     }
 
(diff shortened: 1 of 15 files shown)


Prox/Prox/PlaceDetails/PlaceDetailsCardViewController.swift
```diff
@@ -187,6 +195,8 @@ class PlaceDetailsCardViewController: UIViewController {
             let url = URL(string: urlString) else { return }
         if !OpenInHelper.open(url: url) {
             print("lol unable to open web address")
+        } else {
+            Analytics.logEvent(event: AnalyticsEvent.EVENT_BANNER_LINK, params: [:])
```
fluffyemily: This will only work if there is a URL attached to the event. Is there value in logging that people attempted to interact with an event that had no URL?

Prox/Prox/PlaceDetails/PlaceDetailsDescriptionView.swift
```diff
@@ -55,6 +56,8 @@ class PlaceDetailsDescriptionView: UIView {
          icon: UIImage?,
          horizontalMargin: CGFloat) {
         self.horizontalMargin = horizontalMargin
+        toggleEventType = labelText == "Yelp top review" ? AnalyticsEvent.YELP_TOGGLE : AnalyticsEvent.WIKIPEDIA_TOGGLE
```
fluffyemily: This should not be in the `PlaceDetailsDescriptionView`. It's too fragile. If we were to change the text of the label in `PlaceDetailsCardView` without changing the test here then this analytic wouldn't send.

I suggest adding a variable of `var analyticsEvent: AnalyticsEvent` and set it when we create the description view insode `PlaceDetailsCardView` to the correct value, then use that value instead.

fluffyemily:
Also, APIKeys.plist needs to be uploaded into Buddybuild's secure files https://dashboard.buddybuild.com/apps/57f51f8bcaa70a01005412ee/settings/secure-files else the builds will break

fluffyemily:
ping @liuche. Can you make the small change I requested and merge this please?

liuche:
@fluffyemily pushed the rest of the events in the following commits, and I can merge them all at once.

Issue #261, opened 2016-11-23
[startup crash] NSInvalidArgumentException format:@"Precision must be less than %d!"


mcomella:
This occurs on a fresh install with the latest upstream/master: f42c9725f05deaec730a7d5f436d3d1d24955b52

The code throws in `GFGeoHash initWithLocation:precision:` where the precision parameter is greater than expected.

mcomella:
In GFGeoHashQuery.m, the precision comes from:
```objc
NSUInteger bits = [GFGeoHashQuery bitsForRegion:region];
NSUInteger geoHashPrecision = ((bits-1)/BITS_PER_GEOHASH_CHAR)+1;
```

Region comes from the function declaration: `queriesForRegion:(MKCoordinateRegion)region`

mcomella:
Same file:
```objc
+ (NSSet *)queriesForLocation:(CLLocationCoordinate2D)center radius:(double)radius
{
    CLLocationDegrees latitudeDelta = radius/METERS_PER_DEGREE_LATITUDE;
    CLLocationDegrees latitudeNorth = fmin(90, center.latitude + latitudeDelta);
    CLLocationDegrees latitudeSouth = fmax(-90, center.latitude - latitudeDelta);
    CLLocationDegrees longitudeDeltaNorth = [GFGeoHashQuery meters:radius toLongitudeDegreesAtLatitude:latitudeNorth];
    CLLocationDegrees longitudeDeltaSouth = [GFGeoHashQuery meters:radius toLongitudeDegreesAtLatitude:latitudeSouth];
    CLLocationDegrees longitudeDelta = fmax(longitudeDeltaNorth, longitudeDeltaSouth);
    MKCoordinateRegion region = MKCoordinateRegionMake(center, MKCoordinateSpanMake(latitudeDelta*2, longitudeDelta*2));
    return [GFGeoHashQuery queriesForRegion:region];
}
```

Notably, the radius arg is 0.

mcomella:
I changed `EventsProvider.radius` and `PlacesProvider.radius` to return 27 and I no longer crash.

It seems the firebase remote config is returning 0 (even though the value in the firebase is not 0). Thoughts:
* Remote config doesn't get deleted when an app is removed so I have a stale value
* Remote config sets its own default values when values are missing

This is concerning for our actual release! :\

mcomella:
[...]

PR #263, opened 2016-11-23
Revert "Closes #228 – Harden RemoteConfig to prevent catastrophic fai…

### LINKED ISSUE #228: Harden Remote Config access to reduce catastrophic failure

mcomella:
…lure."

This reverts commit f42c9725f05deaec730a7d5f436d3d1d24955b52 which caused #261.

### DIFF
--- a/Prox/Prox/AppDelegate.swift
+++ b/Prox/Prox/AppDelegate.swift
@@ -18,9 +18,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     private lazy var remoteConfigCacheExpiration: TimeInterval = {
         if AppConstants.isDebug {
             // Refresh the config if it hasn't been refreshed in 60 seconds.
-            return 0.0
+            return 60
         }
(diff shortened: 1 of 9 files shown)


liuche:
mcomella says tests pass, even though Buddybuild doesn't build on this branch

PR #266, opened 2016-11-23
Closes #261 – Parse the strings ourselves instead of relying on RemoteConfig.

### LINKED ISSUE #261: [startup crash] NSInvalidArgumentException format:@"Precision must be less than %d!"

jhugman:
The API for RemoteConfig is a little bit funky.

`remoteConfig[key]` will give you back a value object, even if the key doesn't exist.

Worse still, the value object `FIRRemoteConfigValue` will silently fail if the value isn't coercible to the type that we want, or if the key doesn't exist.

So, now: get the `stringValue` out, try and parse it ourselves; if that fails, look for the plist value (which should be correct, but may not be).

If that fails, the fallback further to something a developer typed in to a swift file.


Prox/Prox/AppDelegate.swift
```diff
@@ -73,7 +73,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     private func setupRemoteConfig() {
         let remoteConfig = FIRRemoteConfig.remoteConfig()
-        remoteConfig.configSettings = FIRRemoteConfigSettings(developerModeEnabled: AppConstants.BuildChannel != AppBuildChannel.Release)!
+        let isDeveloperMode = AppConstants.isDebug || AppConstants.MOZ_LOCATION_FAKING
+        remoteConfig.configSettings = FIRRemoteConfigSettings(developerModeEnabled: isDeveloperMode)!
```
mcomella: What is developer mode? Are we sure we want it on the "Prox Kona" builds?

Prox/Prox/AppDelegate.swift
```diff
@@ -82,21 +83,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         remoteConfig.fetch(withExpirationDuration: remoteConfigCacheExpiration) { status, err in
             if status == FIRRemoteConfigFetchStatus.success {
-                print("RemoteConfig fetched")
+                NSLog("RemoteConfig fetched")
```
mcomella: What's the benefit to `NSLog`?

Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -87,16 +87,34 @@ class RemoteConfigProperty<T> {
...
+            }
         }
-        if let rcv = remoteConfig.defaultValue(forKey: key, namespace: nil),
-            let value = convert(rcv) {
-            print("RemoteConfigKeys: Default value for \(key) from embedded plist")
+
+        // TODO: save the last successful value in UserDefaults.standard, then fall back to it here,
+        // instead of falling back to what's in the plist.
+
+        // Check to see if RemoteConfigDefaults.plist has a parseable default
```
mcomella: nit: this is slightly confusing. My understanding is that `remoteConfig[key]` will return a plist value if we call `setDefaultsFromPlistFile` before accessing the remote config.

Thus, its confusing because the comment "check to see if *.plist has a... default" because I'd think we'd already checked the default value in the original `remoteConfig[key]`.

To be clear, I understand what is happening and that this is a fallback: if `remoteConfig[key]` has been overwritten, we need to call `defaultValue` to get the plist value (though it's redundant in the case we've already checked the plist value).

Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -87,16 +87,34 @@ class RemoteConfigProperty<T> {
...
+
+        // rcv is never nil even if there is no key.
+        // rcv.numberValue? never fails either (in light of an invalid number, it defaults to zero).
+        // So, we get the string value, and try parsing it ourselves.
         let rcv = remoteConfig[key]
-        if let value = convert(rcv) {
-            return value
+        if let string = rcv.stringValue {
+            if let value = convert(string) {
+                return value
```
mcomella: Since I assume the empty string is the default value, this will always pass for Strings so we'll never get the type-checked default values of `RemoteConfigPropertyString` – you should remove the default property for Strings and demanded they have plist values.

Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -87,16 +87,34 @@ class RemoteConfigProperty<T> {
...
+            } else if string != "" {
+                // i.e. no default exists, and no value is specified.
+                NSLog("RemoteConfigKeys: Existing value for \(key) is not a valid \(type(of: defaultValue))")
+            }
         }
-        if let rcv = remoteConfig.defaultValue(forKey: key, namespace: nil),
-            let value = convert(rcv) {
-            print("RemoteConfigKeys: Default value for \(key) from embedded plist")
+
+        // TODO: save the last successful value in UserDefaults.standard, then fall back to it here,
```
mcomella: nit: I'm a bit hesitant of writing a cache over Firebase's cache.

Prox/Prox/RemoteConfigKeys.swift
jhugman: Added a nil check to `RemoteConfigPropertyString.convert`.

Prox/Prox/RemoteConfigKeys.swift
jhugman: yes, me too.

Comment added:

        // This could lead to all sorts of caching related bugs,
        // but would ameliorate the jarring application of a way out of date default.

Prox/Prox/RemoteConfigKeys.swift
jhugman: I've added more comments.

TBC: I really don't want to be using the plist. It doesn't help with type info, has terrible failure conditions.

We should *not* use the plist, though there are compelling cases (very long strings) when we could use them. If I started from scratch, I would not support the plist for defaults.

jhugman:
Merged.

PR #272, opened 2016-11-24
170 / 247 - notification fixes

fluffyemily:
This does the following epic number of things.

Sorry.

- cleans up the notification send criteria logic
- implements messaging according to whether an event is upcoming, about to start, ongoing or about to end for presentation on place cards
- implements messaging according to whether an event is upcoming, ongoing or ending for notifications
- changed the way we remember which notifications we've already sent and ensure that writing of the cache to defaults in the background is done on a background thread
- added coordinates to the event so that we no longer need places when sending event notifications, reducing the amount of battery we use when sending notifications
- add notifications for events that are not within traveling distance at point of sending to the sent notifications so we don't have to calculate travel times again if we know we are going to discard it
- Remove calculation of whether an event is available to be sent from the `EventsProvider` to the `Event`
- Clean up presentation of events on place cards so card reuse doesn't lead to a blue flash as event card is removed


Prox/Prox/Data/TravelTimesProvider.swift
```diff
@@ -8,6 +8,8 @@ import MapKit
 
 struct TravelTimesProvider {
 
+    static fileprivate let travelTimePadding: Double = 10 * 60
```
jhugman: Nit: what unit is this? Comment, put the unit in the name or use the type system.

Prox/Prox/Models/Event.swift
```diff
@@ -64,23 +64,41 @@ class Event {
...
                 print("lol dropping event: missing data, id, placeId, description, start time \(data.value)")
                 return nil
         }
 
+        let eventfulDateFormatter = DateFormatter()
+        eventfulDateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss"
+
+        let gCalDateFormatter = DateFormatter()
+        gCalDateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssz"
+
```
jhugman: What? Is there really a difference? This should be fixed on the server, not in the app.

Prox/Prox/Models/Event.swift
```diff
@@ -64,23 +64,41 @@ class Event {
...
+        var localStartTime: Date
+
+        if let startTime = eventfulDateFormatter.date(from: localStartTimeString) {
+            localStartTime = startTime
+        } else if let startTime = gCalDateFormatter.date(from: localStartTimeString) {
+            localStartTime = startTime
+        } else {
```
jhugman: Can we detect from `localStartTimeString` which `DateFormatter()` is needed?

Prox/Prox/Models/Event.swift
liuche: I thought I had this fixed, but apparently not.

New server pull-request: https://github.com/mozilla-mobile/prox-server/pull/60

fluffyemily: I'm gonna leave this in here for now, but add a comment that we should remove this when the date issue has been sorted out

Prox/Prox/Models/Event.swift
fluffyemily: The only way I can figure that out is to match the date format which `DateFormatter` does. Can't think of any other way

Prox/Prox/Models/Event.swift
```diff
@@ -13,82 +14,237 @@ class Event {
...
+
+    private func replaceTimeToStart(string: String) -> String {
+        let now = Date()
+        let timeToEvent = startTime.timeIntervalSince(now)
+        let timeString = timeToEvent.asHoursAndMinutesString()
+        return string.replacingOccurrences(of: "{time_to_start}", with: "\(timeString)")
+    }
+
+    private func replaceTimeToEnd(string: String) -> String {
+        guard let endTime = endTime else { return string}
```
jhugman: So we have "{time_to_end}" in the display string?

Prox/Prox/Models/Event.swift
```diff
@@ -13,82 +14,237 @@ class Event {
...
+    }
+
+    private func replaceStartTime(string: String) -> String {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "h:mm a"
+        return string.replacingOccurrences(of: "{start_time}", with: formatter.string(from: startTime))
+    }
+
+    private func replaceEndTime(string: String) -> String {
+        guard let endTime = endTime else { return string}
```
jhugman: So we have "{end_time}" in the display string?

Prox/Prox/Utilities/EventNotificationsManager.swift
```diff
@@ -75,7 +78,7 @@ class EventNotificationsManager {
     }
 
     func persistNotificationCache() {
-        UserDefaults.standard.set(sentNotifications, forKey: sentNotificationDictKey)
+        UserDefaults.standard.set([String(Date().timeIntervalSinceReferenceDate): Array(sentNotifications)], forKey: sentNotificationDictKey)
```
jhugman: Should you be `synchronizing` your `UserDefaults`.

Prox/ProxTests/EventTests.swift
```diff
@@ -0,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import XCTest
+@testable import Prox
+
+class EventTests: XCTestCase {
+
+    let oneHour: Double = 60 * 60
```
jhugman: nit: this should be a constant outside of the class definition.

Issue #234, opened 2016-11-21
Spinner and show on map shown at the same time


kbrosnan:
I am seeing the view on map and the spinner at the same time. Not sure how I got into this state.

![screen shot 2016-11-21 at 11 25 48 am](https://cloud.githubusercontent.com/assets/250273/20498977/3806facc-afe4-11e6-8514-fd6c2712f841.png)

mcomella:
Making robustess changes for travel times in #185 – hopefully I'll be able to close this with that patch.

Issue #219, opened 2016-11-19
Map view image & travel times: loading & error states


mcomella:
@vazquez In map view, we make async requests for images and travel times – we'll need to add a placeholder while we wait for these to load. fwiw, in practice with a fast connection, we won't see this, except for travel times if we're not able to solve #184.

Then we need the error state, if the request fails.

vazquez:
### Design

- We have an image placeholder while we wait to retrieve images.
- While we wait for travel times, we will simply show a `Locating...` text string. If we cannot retrieve travel times, we substitute the text string with `Travel time unknown`

![image loading](https://cloud.githubusercontent.com/assets/1119821/20489993/83560f0e-afda-11e6-89d1-585b987ccc21.png)


### Assets

The image loading assets is in the Assets folder. https://www.dropbox.com/sh/b30k7wfqerbbb7f/AACdQhl1QKr0MnQyly6KGngSa?dl=0

`image-loading.png`
`image-loading@2x.png`
`image-loading@3x.png`

fluffyemily:
@vazquez - does this mean you want us to remove the spinner that we
currently display while calculating travel times and replace it with the
word "Locating"?

vazquez:
@fluffyemily The spinner only shows in Summary Card view though, right? We don't have a placeholder for Map View, from what I gather from @mcomella.

mcomella:
> @fluffyemily The spinner only shows in Summary Card view though, right? We don't have a placeholder for Map View, from what I gather from @mcomella.

This is correct.

fluffyemily:
Ah yeah. Sorry. Didn't read the title clearly!

antlam:
@mcomella if "Travel time unknown" how does the user "Try again"? or what can they do about it?

mcomella:
> @mcomella if "Travel time unknown" how does the user "Try again"? or what can they do about it?

The reason we get "Travel time unknown" is that Apple rate-limits us and does not provide the travel time. We retry the next time the card is shown (though we may still be rate-limited).

Speaking with @vazquez, the suggestion of removing the error state and keeping "Locating..." makes sense with the back-end.

Issue #185, opened 2016-11-16
Travel times & images can get out of sync in Place Carousel


mcomella:
Travel times in the map view are not always accurate (see attached photo) – I think this is a re-use problem where we wait for the results of the travel time. Note: I was swiping quickly.

Clicking into the details view will show the correct travel time but closing out of the details view will show the incorrect time in the carousel still.

![screen shot 2016-11-15 at 17 29 53](https://cloud.githubusercontent.com/assets/759372/20331398/e549eff8-ab59-11e6-9a83-7846e4927c90.png)

mcomella:
As per https://github.com/mozilla-mobile/prox/issues/184#issuecomment-260828204, we reload the travel times for each place that's shown. If we can't update the travel times (e.g. rate limited in #184), I believe we'll just display the old value.

mcomella:
I realized this is caused by:
1. not niling the previous value when the cell is re-used
1. making asynchronous requests on cells that can be re-used (when quickly scrolling, the cell is reused so the request starts, the cell is re-used again, the request finishes and updates the UI, making it out of date)

2 also affects images.

Issue #184, opened 2016-11-16
Travel times don't load when swiping quickly (rate limited?)


mcomella:
Some travel times don't load: I assume we're getting rate-limited when swiping quickly through a full carousel (~30 items).

Question: Are we querying once for place or querying multiple times for each place? I don't think we wait for an in-flight query to finish before requesting another one and I think we have an issue with loading the places data more than once, compounding the issue.

mcomella:
It looks like we reload the travel times each time a cell is shown in the map view (PlaceCarousel#91):
```swift
        if let location = locationProvider?.getCurrentLocation() {
            place.travelTimes(fromLocation: location, withCallback: { travelTimes in
                self.setTravelTimes(travelTimes: travelTimes, forCell: cell)
            })
        }
```

fluffyemily:
Might be worth storing travel times by location, so we update our travel times if we have moved a certain distance from the location we were in since we calculated the last travel time for this location, but don't have to do it every time if we haven't moved far since last time.

PR #275, opened 2016-11-24
Fix up travel times

mcomella:
#184 #185 #219 #234 

Important notes:
* You should read the extended commit summaries – I explain some caveats in there
* One problem we still face is, if the user swipes through the carousel quickly and clicks into the detail view, they may see the loading spinner for seemingly forever. This means a user doesn't see "View on map" and they can't get directions. I think this is because the `TravelTimesProvider` methods don't call the callback sometimes when there is a timeout. We should add a timeout (either by figuring out what's up with `TravelTimesProvider` or adding one to the deferred) where we replace a loading spinner with "View on Map"
  * We could probably fix ^ by not letting a user swipe so damn fast too :)
  * We should file a follow-up if you think it's an issue (I didn't)
* I was unable to test on device which might let me wreak more havoc on the UI so maybe that would make this patch series less effective.


Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -42,6 +42,9 @@ class RemoteConfigKeys {
     // This is measured in minutes. The default value is 1 minute.
     public static let youAreHereWalkingTimeMins = RemoteConfigInt(key: "you_are_here_walking_time_mins", defaultValue: 1)
 
+    // The distance the user must travel before the travel times on a location are reloaded.
+    public static let travelTimeExpirationDistance = RemoteConfigDouble(key: "travel_time_expiration_distance", defaultValue: 0.2) // default: ~2.5min walk.
```
fluffyemily: can we have the unit of measurement in the key name here, like with the other keys? this is so it makes it easy to update without having to look at the code.  So, `travel_time_expiration_distance_meters` or whatever the unit is.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -95,12 +95,21 @@ struct PlaceUtilities {
             return
         }
 
-        // TODO: must cancel in-flight, or prevent updates to existing UI.
         let travelTimesResult = place.travelTimes(fromLocation: location)
         if !travelTimesResult.isFilled {
             view.setTravelTimesUIIsLoading(true)
         }
+
+        let idAtCallTime = place.id
```
fluffyemily: So, every object in the UIView heirarchy has an `restorationIdentifier` property that can be used to attach identifiers to views for the purposes or reuse or restoration. You could use that, so 
`(view as? UIView)?.restorationIdentifier = place.id` for example

PR #293, opened 2016-11-25
Issue 60 - Added crossfade animation when setting the background image of the detail view

sleroux:
Added a crossfade when setting the background image in the detail view controller whenever the user moves either the cards or the pages.

### DIFF
--- a/Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
+++ b/Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
@@ -198,6 +198,17 @@ class PlaceDetailViewController: UIViewController {
     fileprivate func setBackgroundImage(toPhotoAtURL photoURLString: String?) {
         if let imageURLString = photoURLString,
             let imageURL = URL(string: imageURLString) {
+
+            let imageRequest = URLRequest(url: imageURL)
+            let cachedImage = UIImageView.sharedImageDownloader().imageCache?.imageforRequest(imageRequest, withAdditionalIdentifier: nil)
+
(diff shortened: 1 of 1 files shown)


Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -198,6 +198,17 @@ class PlaceDetailViewController: UIViewController {
...
+            let cachedImage = UIImageView.sharedImageDownloader().imageCache?.imageforRequest(imageRequest, withAdditionalIdentifier: nil)
+
+            // Perform a cross fade between the existing/new images
+            let crossFade = CABasicAnimation(keyPath: "contents")
+            crossFade.duration = 0.4
+            crossFade.fromValue = backgroundImage.image?.cgImage
+            crossFade.toValue = cachedImage
+            backgroundImage.layer.add(crossFade, forKey: "animateContents")
+
             self.backgroundImage.setImageWith(imageURL)
```
fluffyemily: if we've already got the cached image, do we still need to use `backgroundImage.setImageWith(imageURL)` will this not just download the image again (although I suppose we'll just get it from the cache). But perhaps `self.backgroundImage.image = cachedImage`?

sleroux: `cachedImage` will usually return nil when navigating to new images since we haven't fetched them yet so in these cases we'll want to call `setImageWith(url:)` to fetch+load them. If we've already fetched them, we'll just hit the cache so it's the same assigning the image directly.

PR #296, opened 2016-11-25
Issue 279 - Fix map centering by moving it to the draw call instead of UIButton image

sleroux:


fluffyemily:
I am not sure how, but this seems to have added a shadow to the map image:

Master:

screen shot 2016-11-25 at 18 14 29

This branch:

screen shot 2016-11-25 at 18 14 01

Issue #174, opened 2016-11-15
Sort order does not place "You're here" first


mcomella:
I opened the app today and "The Bay Club at Waikoloa Beach Resort" was the first card shown despite having a 2min walking time. "Kamuela Provision Company", which has normally been first, is second and still has "You're here"

Questions:
* Can multiple cards have "You're here"? Should they be able to?
* How do we sort cards? By distance or travel times?

mcomella:
> Should multiple cards be able to have "You're here"?

@antlam, what do you think?

mcomella:
fwiw, it's slightly easier for us if they can.

antlam:
@mcomella having multiple cards show "You're here" is interesting. But it can create a lot of confusion for our users right now. I think we should avoid this for Kona ATM.

> Can multiple cards have "You're here"? Should they be able to?

Not right now. This should always be the first card.

> How do we sort cards? By distance or travel times?

Travel times

mcomella:
@antlam When we sort by travel times, should we sort by their walking travel times or overall travel time? I'm afraid with the latter, we'd get roughly one "3min walk", a lot of "3min drive", then a regular distribution of "10min walk/drive".

antlam:
@mcomella I was talking to Brad and Sebastian about this in Taipei. 

Brad suggested an approach like this, which I think makes a lot of sense: 

1. Assume an inherent additional "lag time/cost" of 15 minutes to all Driving times
2. Sort all places in the backend by this "Total travel time" (Driving time + 15min) 
3. Display using "Walking time" IF a place is less than 15 minutes walking time away, display this in "# min walk".
4. Display using "Driving time" everywhere else, display this in "# min drive". NOT "(# + 15) min drive".

Note: We will still have a scenario where we see "1 min walk", "3 min walk", "3 min drive", "4 min drive", but I don't see why that's that's a huge issue.

Does this make sense to you? This way, the first couple results should still be "Walking times" and those would be things you could reach immediately within 15 minutes. Anything further, we'd display Driving times.

I based 15 minutes on how far an average person walks a day (from a quick googling - https://www.reference.com/health/many-miles-average-person-walk-per-day-d9d3aa8ef93e9ac3) and how long that would approximately take.

antlam:
@mcomella looking back, I'm not sure if some of that ^ may be out of scope for this issue. Feel free to open another issue if it is. Just copy over what we're discussing above.

mcomella:
[...]

PR #299, opened 2016-11-25
Issue 174 - sort by travel times

fluffyemily:
#174 

I think I worked through all the new deferred stuff to get travel times sorting working


Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
+import Deferred
 
 struct PlaceUtilities {
```
mcomella: nit: I assume `PlaceUtilities.sort(...:byDistanceFromLocation` is unused now – shall we remove it?

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -21,6 +22,43 @@ struct PlaceUtilities {
...
+        }
+    }
+
+    private static func getTravelTimes(forPlaces places: [Place], fromLocation location: CLLocation) -> Future<[DatabaseResult<TravelTimes>]> {
+        var placesWithTimes = [Deferred<DatabaseResult<TravelTimes>>]()
+        for place in places {
+            let placeDeferred = place.travelTimes(fromLocation: location)
+            placesWithTimes.append(placeDeferred)
+        }
+        return placesWithTimes.allFilled()
```
mcomella: Does this always complete? I was having issues with the travel time UI occasionally only showing a spinner when scrolling quickly – I don't know if this was a UI bug or a logic bug, but I vaguely remember it being the `TravelTimesProvider` method not returning.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -21,6 +22,43 @@ struct PlaceUtilities {
...
+                guard ascending else {
+                    return placeAETA > placeBETA
+                }
+                return placeAETA < placeBETA
+            }
+            completion(sortedPlaces)
+        }
+    }
+
+    private static func getTravelTimes(forPlaces places: [Place], fromLocation location: CLLocation) -> Future<[DatabaseResult<TravelTimes>]> {
```
mcomella: nit: Not necessary to change and I'm not sure if it's your style, but you could replace this method with:
```swift
let placesWithTimes = places.map { place in place.travelTimes(fromLocation: location }
return placesWithAllTimes.allFilled()
```

which I find easier to quickly understand the intent (since a for loop can do anything but `map` has a specific result).

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -21,6 +22,43 @@ struct PlaceUtilities {
         }
     }
 
+    static func sort(places: [Place], byTravelTimeFromLocation location: CLLocation, ascending: Bool = true, completion: @escaping ([Place]) -> ()) {
+        PlaceUtilities.getTravelTimes(forPlaces: places, fromLocation: location).upon { result in
+            let sortedPlaces = places.sorted { (placeA, placeB) -> Bool in
+                guard let placeATravelTimes = PlaceUtilities.lastTravelTimes(forPlace: placeA),
+                    let placeBTravelTimes = PlaceUtilities.lastTravelTimes(forPlace: placeB) else {
+                        return false
```
mcomella: This logic is inconsistent – we'd put a null A after B but not put a null B before an A. Where do we want to place items without travel times? In the back? Also, consider what we should do with ascending in this case.

If we want them in the back, maybe we should:
```swift
let placeATravelsTimes = ...
...
let placeAETA = placeATravelTimes?.getShortestTravelTime() ?? MAX_VALUE
```

Keep in mind we'd have to rotate for ascending.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -21,6 +22,43 @@ struct PlaceUtilities {
...
+                }
+
+                let placeAETA = placeATravelTimes.getShortestTravelTime()
+                let placeBETA = placeBTravelTimes.getShortestTravelTime()
+
+                guard ascending else {
```
mcomella: nit: I might just be a Swift noob but this is confusing as a `guard`.

(But I might also argue that having what amounts to an `if` statement with reverse semantics can also confusing! :P But I like the intent of showing early returns)

PR #301, opened 2016-11-25
Listen for the foregrounded notification and ensure that places are f…

fluffyemily:
…etched when location fetch comes back

Ensure that we always fetch new location when app opens


fluffyemily:
[Polish: As a user I wish to be able to launch the app in a new place, and eventually see venues so I can see venues in unfamiliar places.](https://trello.com/c/xe4W9mwt/227-polish-as-a-user-i-wish-to-be-able-to-launch-the-app-in-a-new-place-and-eventually-see-venues-so-i-can-see-venues-in-unfamiliar-)

Issue #123, opened 2016-11-08
Handle multiple open hours intervals for a Place


mcomella:
If a location is open multiple times (e.g. open for lunch and dinner but closed in between), for simplicity, we currently only handle one of these intervals [1]. This issue is to make us take into account all intervals.

FYI @antlam @vazquez.

[1]: the last one in the array, which is likely sorted from the server, so it'll be the latest opening time

PR #313, opened 2016-11-28
#123 multiple open hours

fluffyemily:
- Turn OpenHours into an collection of days with multiple open periods.
- Update isOpen to correctly determine whether a place is open when it has multiple open periods
- Update fetching of opening and closing time strings to return correct value when it has multiple open periods
- Update tests to reflect changes


Prox/Prox/Models/Place.swift
```diff
@@ -388,58 +388,79 @@ struct OpenHours {
...
-            let closingTime = getClosingTime(forDate: time) else {
+        guard let allOpeningTimes = getOpeningTimes(forDate: time) else {
             return false
         }
 
-        // Is the current time before the opening time and the venue closed after midnight the previous day?
-        if time < openingTime,
-            let yesterdaysClosingTime = getClosingTime(forDate: time.addingTimeInterval(-AppConstants.ONE_DAY)),
-            yesterdaysClosingTime < openingTime {
+        guard let _ = getCurrentOpeningPeriod(fromOpeningPeriods: allOpeningTimes, forDate: time) else {
```
mcomella: nit: It's confusing what you're trying to guard against. Perhaps you could clarify with a named variable:
```swift
let isOpen = getCurrentOpeningPeriod... != nil
guard isOpen else {
```

It could also help to return early so I don't have to keep the stored state in my head:
```swift
let isOpen = ...
if isOpen {
    return true
}

// do not open stuff
```

Prox/Prox/Models/Place.swift
```diff
@@ -388,58 +388,79 @@ struct OpenHours {
...
+            // check to see if the current time is before the earliest open time. If it is, it may be before yesterdays closing time too
+            // otherwise we're just not open
+            guard let earliestOpeningPeriod = getEarliestOpeningPeriod(fromOpeningPeriods: allOpeningTimes),
+                time < earliestOpeningPeriod.openTime,
+                let yesterdaysOpeningTimes = getOpeningTimes(forDate: time - AppConstants.ONE_DAY),
+                let _ = getCurrentOpeningPeriod(fromOpeningPeriods: yesterdaysOpeningTimes, forDate: time) else {
```
mcomella: nit: also a little confusing

Prox/Prox/Models/Place.swift
```diff
@@ -451,25 +472,30 @@ struct OpenHours {
...
     func nextOpeningTime(forTime time: Date) -> String? {
-        guard let openingTime = getOpeningTime(forDate: time) else {
-                return nil
-        }
+        guard let allOpeningTimes = getOpeningTimes(forDate: time),
+            let openingTime = getNextOpeningPeriod(fromOpeningPeriods: allOpeningTimes, forDate: time) else {
 
-        if time < openingTime {
-            return timeString(forDate: openingTime)
+                guard let tomorrowsOpeningPeriods = getOpeningTimes(forDate: time + AppConstants.ONE_DAY),
```
mcomella: nit: maybe I'm a noob, but nesting guards is mind-bendingly confusing.

`guard` feels good to me when it's used as an early return for error states (e.g. arguments are unexpectedly nil) but requires mental overhead when used for other situations because I have to take what's written and reverse it.

Prox/Prox/Models/Place.swift
```diff
@@ -451,25 +472,30 @@ struct OpenHours {
     }
 
     func nextOpeningTime(forTime time: Date) -> String? {
-        guard let openingTime = getOpeningTime(forDate: time) else {
-                return nil
-        }
+        guard let allOpeningTimes = getOpeningTimes(forDate: time),
+            let openingTime = getNextOpeningPeriod(fromOpeningPeriods: allOpeningTimes, forDate: time) else {
```
mcomella: Could we simplify this by  appending tomorrow's earliest opening time to `allOpeningTimes` returning `getNetxOpeningPeriod` to that?

Issue #217, opened 2016-11-18
if image is not present do not display carousel


kbrosnan:
Feels rather broken to swipe through a set of blank images
![screen shot 2016-11-18 at 2 01 14 pm](https://cloud.githubusercontent.com/assets/250273/20448171/91e263be-ad97-11e6-82b2-fded2072b098.png)

mcomella:
Hiding the carousel is hard so we don't want to change the view, but consider the following for the sprint. If no images:
* Change carousel to be one dot so you can't swipe through empty images
* Set BG to a different, less obvious color.

antlam:
@vazquez has an error state for this on the Map view. We should just use that here for this sprint.

mcomella:
To be clear, the error state is a placeholder image while we wait for the image to load – this should be trivial with AFNetworking (there is a placeholder arg to load image).

Make sure we cancel any in-flight requests (if load image does not do so already).

vazquez:
The image placeholder has been placed in the Assets folder.

`image-placeholder.png`
`image-placeholder@2x.png`
`image-placeholder@3x.png`

https://www.dropbox.com/sh/b30k7wfqerbbb7f/AACdQhl1QKr0MnQyly6KGngSa?dl=0

This is what the image placeholder should look like:

![home - tropics ale house copy 2](https://cloud.githubusercontent.com/assets/1119821/20765599/fe0c3b24-b700-11e6-8f2b-4e0e0c12d851.png)

mcomella:
Design decision from @vazquez on slack:
* Set the background image to the place holder image of the hotel we used at the start of the project when entering a view with a placeholder

PR #344, opened 2016-11-29
Issue 333 - Fixed potential divide-by-0 crash when image carousel is empty

sleroux:


Prox/Prox/PlaceDetails/PlaceDetailsCardViewController.swift
```diff
@@ -295,6 +295,12 @@ extension PlaceDetailsCardViewController: UIScrollViewDelegate {
 
     fileprivate func didChangePage(scrollView: UIScrollView) {
         let pageSize = imageCarouselCollectionView.bounds.size
+
+        // Unable to change pages if we don't have anything to page!
```
mcomella: nit: This should be a little clearer that "we don't have anything to page" when the image carousel is empty.

sleroux:
> In a more ideal world, perhaps we'd consider explicitly adding a placeholder for the image carousel if it's empty so it doesn't negatively affect the layout (e.g. if we change the layout in the future).

Definitely. I think https://github.com/mozilla-mobile/prox/issues/217 covers this case.

Issue #317, opened 2016-11-28
Responding to a notification while app is doing place search does not bring up event


fluffyemily:
I received an event notification and responded to it. Waking the app kicked off a place search, which took precendence over the event place display. I waited until the place search was complete (no results) and still the place for the event did not appear.

We should always be able to fetch a place for which we have fired an event notification.

fluffyemily:
This seems to be because we have a number of events in the DB without associated places.

fluffyemily:
Also, moving from place id as event id to a separate event ID means that we never find a place and therefore never open the event. This is Not Good.

PR #346, opened 2016-11-29
Issue #317 - notification response failure

fluffyemily:
So, 2 notification issues

1 - we were not sending notifications in advance of events due to some stupid date idiocy when getting the notification text
2 - we were not responding to notifications because we changed place and event ids to be different but I didn't update the place event fetching code to reflect that.

So, basically, I was an idiot at least twice and notifications paid the price. Sorry.


Prox/Prox/Data/PlacesController.swift
```diff
@@ -44,12 +44,12 @@ class PlacesProvider {
...
     }
 
-    func placeWithEvent(forKey key: String, callback: @escaping (Place?) -> ()) {
+    func place(withKey placeKey: String, forEventWithKey eventKey: String, callback: @escaping (Place?) -> ()) {
         let eventProvider = EventsProvider()
         var placeWithEvent: Place?
         var eventForPlace: Event?
         let lock = NSLock()
-        place(forKey: key) { place in
+        place(forKey: placeKey) { place in
```
mcomella: If the place or event cannot be found, we never call the callback.

I think it's unlikely to happen since we wouldn't have received an event with a bad ID or an invalid placeID so it's probably not worth fixing, but perhaps add a comment warning about this possibility.

Prox/Prox/Models/Event.swift
```diff
@@ -202,8 +202,8 @@ class Event {
         return currentTime > lastNotificationTime - Event.eventStartNotificationInterval && currentTime < lastNotificationTime
     }
 
-    private func isFutureEvent(eventStartTime: Date, currentTime: Date) -> Bool {
```
mcomella: This method is called by: `internal func shouldShowEvent(withStartTime startTime: Date, endTime: Date?, ...`

It's fragile that `isFutureEvent` does not use the times passed in with `shouldShowEvent` (which is called with the instance's `startTime` and `endTime`).

Recommendations:
* (simpler, but perhaps misleading that it doesn't use instance vars) Revert `isFutureEvent` to take these parameters
* Remove the startTime/endTime parameters from `shouldShowEvent`. I think they're there to make for easier testing so you'd have to add a test helper method like `getEvent(withStartTime:withEndTime) -> Event` or something and fix up the tests.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -305,10 +305,21 @@ class PlaceCarouselViewController: UIViewController {
         self.present(alertController, animated: true)
     }
 
-    func openPlaceForEvent(withKey key: String) {
-        placesProvider.placeWithEvent(forKey: key) { place in
+
+    fileprivate func openPlace(_ place: Place) {
```
mcomella: nit: This appears to be un-used.

PR #348, opened 2016-11-29
No Issue - Reduced vertical spacing between faded out cards

sleroux:
Small UX tweak to the faded out cards


mcomella:
> If reduced vertical spacing is what UX wants, then reduced vertical spacing is what they shall have!

Important note: this logic does not to apply to everything UX wants. 😜

PR #349, opened 2016-11-29
Closes #321: Disable Firebase persistence.

### LINKED ISSUE #321: Non-cached locations may show fewer results due to Firebase persistence
Ricardo and I were getting different search results despite having the same search radius (we override the Remote Config value in code) and same commit checkout 9a60154b18c685cb53dc9d2a7a9c8ab08027814c.

Notably, I was using a fresh install on the simulator and Ricardo was using his with data from a previous build. Ricardo was receiving fewer places (e.g. ~3 places for 1 km where I received ~20). We both disabled firebase persistence and received the larger place count (~20 places).

---
Since we do a one-time Firebase query (value?), I'd guess Firebase is just retrieving from the cache and not actually contacting the server for new data. I don't know if it'd ever update that data (e.g. we queried multiple times on Ricardo's build) without doing a long-term query that waits for new children to arrive ([FIRDataEventTypeChildAdded?](http://cocoadocs.org/docsets/FirebaseDatabase/3.1.1/Constants/FIRDataEventType.html)).

Firebase persistence **should work for Kona, given that all the data is cached**, however, I largely expect it to return only cached results with new locations.

I think the simple solution is to disable persistence so we don't have to spend time investigating what works and what doesn't work with Firebase.

@jhugman is ^ feasible? Will it cause server problems? Do we perceive too much data use on device?

mcomella:
#321

It appears to prevent users from retrieving new places for a location if they
already have cached values.


mcomella:
Wrong branch.

PR #352, opened 2016-11-29
Issue 351 - Don't show error messaging on loading screen when our PUT request has failed

sleroux:


Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -416,7 +416,6 @@ extension PlaceCarouselViewController: PlacesProviderDelegate {
     }
 
     func placesProvider(_ controller: PlacesProvider, didError error: Error) {
-        showErrorMessageIfNoPlaces()
```
jhugman: I'm not sure if I understand when `didError` should be called.

If we have no network, we shouldn't be able to get to firebase, so will be reliant of cached values.

If we have network, but the server is down, then we should be able to get to firebase.

In that case, I'm not sure we should be doing anything with `didError` and we should definitely not be setting `isUpdating = false`.

Probably add a comment to that effect in `PlacesProvider`.

sleroux: Initially I was a bit confused as to what this network call was doing but it looks like we actually don't care about if it error'ed or succeeded since it's just a ping. We could probably just remove the `didError` protocol method altogether. The only 'error' state we have for fetching places should be handled by `didTimeout`.

PR #360, opened 2016-11-30
Closes #250: Move Remote Config values from FB to client defaults.

### LINKED ISSUE #250: Set final RemoteConfig values as default values
Roughly, the `RemoteConfig` will only be updated on second run so we shouldn't rely on it for our initial values.

For this bug, put the final remote config constants we decide on for shipping at the all-hands as default values locally on the app.

mcomella:
This will ensure the client has the most up-to-date configuration on initial
launch, rather than having to wait for the first config to be pulled down (~30
seconds after launch).

`travel_time_expiration_distance` had an empty value on Firebase so I changed
its value to the client default value.

Additionally, the following keys exist on Firebase, but not on the client:
* event_notification_string_1 & 2
* place_details_event_string_1 & 2
* number_of_event_notification_strings
* number_of_place_details_event_strings
* dummy_number

I have taken no action. (@jhugman, do you think I should remove these from Firebase?)

I did not check for keys that exist on the client that do not exist in
Firebase.


mcomella:
> I have taken no action. (@jhugman, do you think I should remove these from Firebase?)

I think we should so we don't accidentally change values that we intend to do something but actually do nothing.

Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -36,7 +36,7 @@ class RemoteConfigKeys {
 
     // The search radius the app will use to query the Firebase database.
     // This is measured in kilometers. The default value is 4 km.
```
jhugman: Nit: comment is out of date.

jhugman: Nit: would also put a comment for `place_categories_to_hide_csv` to see long default in the plist.

PR #361, opened 2016-11-30
Issue 160 - Display notifications inside app if fired while app in in the foreground

fluffyemily:
Please ignore the first 2 commits - the patch needs rebasing to get rid of them


Prox/Prox/AppDelegate.swift
```diff
@@ -179,7 +184,12 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
                                 willPresent notification: UNNotification,
                                 withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
         // show a badge.
-        completionHandler(UNNotificationPresentationOptions.alert)
+        if let eventKey = notification.request.content.userInfo[notificationEventIDKey] as? String,
+            let placeKey = notification.request.content.userInfo[notificationEventPlaceIDKey] as? String {
+            Analytics.logEvent(event: AnalyticsEvent.EVENT_NOTIFICATION, params: [:])
```
sleroux: I wonder if we want to add a parameter to the flurry event to differentiate between remote/local notifications

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -193,7 +198,12 @@ class PlaceCarouselViewController: UIViewController {
         // apply the constraints
         NSLayoutConstraint.activate(constraints, translatesAutoresizingMaskIntoConstraints: false)
 
-        toggleLoadingUI(loading: true)
+        toggleLoadingUI(loading: shouldFetchPlaces)
```
sleroux: Just a heads up that this will conflict with my loading screen->detail animation patch as I've replaced this method with a `isLoading` property https://github.com/mozilla-mobile/prox/pull/373

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -193,7 +198,12 @@ class PlaceCarouselViewController: UIViewController {
...
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        loadingOverlay.isHidden = false
```
sleroux: IIRC the issue with the non-touchable view was because of the scrolling needed during the transition/cell not found. Do we need this? Might be safer to exclude it if it's not required.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -24,6 +25,8 @@ fileprivate let placesFetchMonitorIdentifier = "PlaceFetchRadiusMonitor"
 
 class PlaceCarouselViewController: UIViewController {
 
+    lazy var placesWithEvents = Set<Place>()
```
sleroux: I noticed that we never assign anything to this - it's always the empty set. Is this needed or is there an assignment missing? If we don't need it we can probably clean up a lot of the changes in the dataSource that was made to accommodate this set.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -330,43 +332,112 @@ class PlaceCarouselViewController: UIViewController {
...
+
+    func fetchPlace(placeKey: String, withEvent eventKey: String, callback: @escaping (Place?) -> ()) {
+        if let placeIndex = places.index(where: { $0.id == placeKey }) {
+            let place = places[placeIndex]
+            if place.events.contains(where: { $0.id == eventKey }) {
+                callback(place)
+            }
+            let eventProvider = EventsProvider()
+            eventProvider.event(forKey: eventKey) { event in
+                callback(place)
```
sleroux: The event is ignored here. Did you need it for anything? Seems weird to get the event then ignore it.

Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -309,6 +311,7 @@ class PlaceDetailViewController: UIViewController {
                         mapButtonBadge.widthAnchor.constraint(greaterThanOrEqualToConstant: 20.0)]
 
         NSLayoutConstraint.activate(constraints, translatesAutoresizingMaskIntoConstraints: false)
+        NSLog("View finished loading")
```
sleroux: Nit: Probably don't need this

Prox/Prox/Widgets/InAppNotificationToast.swift
```diff
@@ -0,0 +1,104 @@
...
+        label.backgroundColor = .clear
+        label.textAlignment = .center
+        return label
+    }()
+
+    fileprivate var presentationTimer: Timer?
+
+    weak var delegate: InAppNotificationToastDelegate?
+
+    fileprivate var viewBottomAnchor: NSLayoutConstraint!
```
sleroux: Bottom anchor isn't used for anything. Also a bit concerning to have these as explicitly unwrapped just in case!

sleroux:
I was able to get a bunch of notifications for events in Toronto just by having the app opened and noticed that 1/2 the time while in the card view the notifications lead me to places and the other 1/2 of the time it failed. It seems that whenever it fails to go to the card, it's because the callback for the `placesProvider.place(withKey:forEventWithKey)` call back was never fired [1]. 

Digging into that method, it looks like we run two concurrent async tasks (fetching the place and fetching the event) and use a lock to mark each callback as a critical section. We call the callback from either one of those tasks so we don't really know where the callback is being called. 

Anyways, not sure if this is just a data issue and I'm looking too much into it but not being able to know where the callback comes from seems concerning.

[1] https://github.com/mozilla-mobile/prox/pull/361/files#diff-28319a278e726e589796867f78f1bfc0R437

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: leftovers from things that I tried to fix the non-interaction of details cards on display. Will remove

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: leftovers from things that I tried to fix the non-interaction of details cards on display. Will remove

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
fluffyemily: leftovers from things that I tried to fix the non-interaction of details cards on display. Will remove

Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
fluffyemily: leftovers from things that I tried to fix the non-interaction of details cards on display. Will remove

fluffyemily:
@sleroux I am concerned because the dev Firebase contains a lot of events without places, which would therefore fail to call the callback, but even once we call the callback as in the fixes I am just about to push, we would still do nothing as the place does not exist.

I used to fetch the place when I fetched the event when fetching the events for notification, but stopped when I realised I could use the lat long inside the event to determine location and on the basis that we shouldn't have an event in the DB without an associated place, but I am wondering whether I should reinstate that as that doesn't seem to be the case...

Prox/Prox/Data/PlacesController.swift
```diff
@@ -48,28 +48,38 @@ class PlacesProvider {
...
+                        callback(nil)
+                    }
                     return
             }
             foundPlace.events.append(event)
             callback(foundPlace)
         }
 
         eventProvider.event(forKey: eventKey) { event in
+            eventReturned = true
```
sleroux: Nit: Might be safer to move this after the lock.lock() call since this variable is being written two from potentially two different threads at different times.

sleroux: Sorry - it's not being written in both places but being written here and read in the other thread so you'll have a race condition.

fluffyemily:
@sleroux I've added a check to ensure we don't send a notification if we haven't got an associated place in the DB

PR #364, opened 2016-11-30
Issue 312 - Resume animation pulsing when the app returns to the foreground

sleroux:
CAAnimations on layers are removed when the app goes into the background [1]. To allow the animation to continue to pulse after backgrounding, the loading widget now listens to the app lifecycle events to explicitly stop/start the animation.

[1] https://forums.developer.apple.com/thread/15796


Prox/Prox/Widgets/LocationLoadingView.swift
```diff
@@ -11,6 +11,7 @@ fileprivate func createCircle(color: CGColor, frame: CGRect) -> CAShapeLayer {
     circle.path = CGPath(ellipseIn: frame, transform: nil)
     circle.fillColor = color
     circle.frame = frame
+    circle.opacity = 0.2
```
sleroux: Changed the default opacity to 0.2 because when we came back to the foreground the blue was dark before starting the animation again. This makes the transition less jolting.

Prox/Prox/Widgets/LocationLoadingView.swift
```diff
@@ -93,19 +94,51 @@ class LocationLoadingView: UIView {
...
         let aAnim = animation(fromState: .shrunk, toState: .enlarged)
         self.circleA.add(aAnim, forKey: "a")
 
         let bAnim = animation(fromState: .enlarged, toState: .shrunk)
         self.circleB.add(bAnim, forKey: "b")
     }
 
-    required init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+    fileprivate func stopPulsing() {
```
sleroux: Not needed but I figured if we have a start we should probably have an explicit stop.

Prox/Prox/Widgets/LocationLoadingView.swift
mcomella: nit: add a comment to that effect?

Prox/Prox/Widgets/LocationLoadingView.swift
mcomella: nit: Add a comment to this effect too! :)

PR #374, opened 2016-11-30
Closes #358 – separate travel times cache from Place instances

### LINKED ISSUE #358: Reduce occurence of missing travel times: times appear to be requested twice
If you add debug statements, on startup, we do one large batch of travel times calculations, then we do another – it appears we do these for the same places. If we didn't make so many requests, we'd see the travel time error states – "Locating..." and "View on Map" – less frequently.

This doesn't seem to be any more important than other things I'm working on so let's let UX reprioritize this.

Debug notes:
* We only set the places list once (the assignment is called twice but we short-circuit since the list is the same).
  * One appears before we show places ("Uh oh! No places found") and another when we initially show the places
* `Place.travelTimes` is called 6 times for a single place ID (once for details and once for map view is expected) – some of these return cached results
  * A fresh result is retrieved for two of these times: `Place.lastTravelTime == null`
    * The memory address of the place changes between invocations
    * Both are called from `PlaceUtilities.sort(...:byTravelTime)`

mcomella:
#358.


Prox/Prox/Models/Place.swift
```diff
@@ -510,3 +519,26 @@ struct OpenHours {
         return OpenHours.calendar.dateComponents(OpenHours.dateComponentsSet, from: date)
     }
 }
+
+// This could be replaced with a thread-safe dictionary, but I didn't see one in the libs. :(
```
fluffyemily: You could use NSCache which is thread safe, but that might not allow a tuple as a value

Issue #356, opened 2016-11-30
alam@mozilla.com on build #529: Bottom isn't at 16 px padding


sarentz:
Feedback from alam@mozilla.com : Bottom isn't at 16 px padding 

<img src="https://s3-us-west-2.amazonaws.com/buddybuild-screenshots/57f51f8bcaa70a01005412ee/583dc8ca7017c20100bd23f4/7757558d-0fdd-421d-8b47-e9115e0c2968.jpg" width="33%" height="33%" /><table><tr><td>Created</td><td>Wed Nov 30 2016 00:04:44 GMT+0000 (UTC)</td></tr><tr><td>CFBundleShortVersion</td><td>1.0</td></tr><tr><td>CFBundleVersion</td><td>1</td></tr><tr><td>App uptime</td><td>376.5229</td></tr><tr><td>Build</td><td>529</td></tr><tr><td>Device type</td><td>iPhone 7</td></tr><tr><td>Device name</td><td>Antlam's 📱 7</td></tr><tr><td>Screen size</td><td>375</td></tr><tr><td>Screen size</td><td>375px by 667px</td></tr><tr><td>Battery</td><td>65% Charging</td></tr><tr><td>Memory free</td><td>146 MB / 1510 MB</td></tr><tr><td>Disk free</td><td>74160 MB / 116301 MB</td></tr><tr><td>Network IP</td><td>0.0.0.0</td></tr></table>

 [Link to buddybuild feedback from build 529](https://dashboard.buddybuild.com/apps/57f51f8bcaa70a01005412ee/feedback?fid=583e179c9a5a8601009ca86c&bnum=529)

mcomella:
I think I fixed this in ^. Here's a post patch screenshot:

![simulator screen shot nov 30 2016 18 45 53](https://cloud.githubusercontent.com/assets/759372/20780065/59fa085e-b72d-11e6-94fd-63801e85dc6e.png)

mcomella:
Merged #379.

Issue #318, opened 2016-11-28
alam@mozilla.com on build #484: Spacing around Yelp info is incorrect


sarentz:
Feedback from alam@mozilla.com : Spacing around Yelp info is incorrect

<img src="https://s3-us-west-2.amazonaws.com/buddybuild-screenshots/57f51f8bcaa70a01005412ee/583c0eaefc26cf010072ad21/0dac8b30-56ca-4eda-aeec-91569ea51cf8.jpg" width="33%" height="33%" /><table><tr><td>Created</td><td>Mon Nov 28 2016 19:23:45 GMT+0000 (UTC)</td></tr><tr><td>CFBundleShortVersion</td><td>1.0</td></tr><tr><td>CFBundleVersion</td><td>1</td></tr><tr><td>App uptime</td><td>118.9225</td></tr><tr><td>Build</td><td>484</td></tr><tr><td>Device type</td><td>iPhone 7</td></tr><tr><td>Device name</td><td>Antlam's 📱 7</td></tr><tr><td>Screen size</td><td>375</td></tr><tr><td>Screen size</td><td>375px by 667px</td></tr><tr><td>Battery</td><td>100% Charging</td></tr><tr><td>Memory free</td><td>109 MB / 1802 MB</td></tr><tr><td>Disk free</td><td>74636 MB / 116301 MB</td></tr><tr><td>Network IP</td><td>10.244.37.172</td></tr></table>

 [Link to buddybuild feedback from build 484](https://dashboard.buddybuild.com/apps/57f51f8bcaa70a01005412ee/feedback?fid=583c844177ec820100049db1&bnum=484)

mcomella:
As you swipe through cards, the spacing appears to change. Perhaps we're accidentally incrementing the spacing (or, if we didn't disable it as I thought we did, card re-use is causing problems).

antlam:
In the previous spec, the padding on the left and right of the content here "The food here is really good..." (in the screenshot above) was set at 16px. We'd like to increase this to 24px to match the title of the location.

@mcomella can we add that to this issue since this is about fixing the padding in general?

mcomella:
> @mcomella can we add that to this issue since this is about fixing the padding in general?

Done.

PR #379, opened 2016-12-01
#318: Fix spacing on expandable views

mcomella:
#318 

And other related changes.


mcomella:
I think this will also close #356.

PR #388, opened 2016-12-01
Issue 357 - Add horizontal line as separate view in header

sleroux:
`PlaceCarouselHeaderView` was actually a subclass of the `HorizontalLineView` which sets the background color to clear in it's draw call which the system will call when it needs to which would make this issue appear 'randomly'. I've changed `PlaceCarouselheaderView` to inherit from the basic `UIView` and add the line as it's own view. I also cleaned up the line logic to just draw a line instead of a UIBezierPath. Lastly, I noticed that since the sun view was on top of the header view, the shadow from the sun view was appearing on the top slightly. I've moved the header view on top so this shadow is hidden.


Prox/Prox/PlaceCarousel/PlaceCarouselHeaderView.swift
```diff
@@ -13,27 +13,30 @@ class PlaceCarouselHeaderView: HorizontalLineView {
...
-        
-        self.setupViews()
-    }
 
-    // setting up the constraints for the views
-    func setupViews() {
-        color = Colors.carouselViewHeaderHorizontalLine
+        line.startX = 20.0
+        line.color = Colors.carouselViewHeaderHorizontalLine
+        line.backgroundColor = .clear
```
mcomella: nit: it could be a little confusing that you have to set the background to .clear on HorizontalLineView if you want to use it as a line each time.

Add a comment to the class? Or set .clear as a default value somewhere.

Prox/Prox/PlaceCarousel/PlaceCarouselHeaderView.swift
```diff
@@ -13,27 +13,30 @@ class PlaceCarouselHeaderView: HorizontalLineView {
...
         addSubview(numberOfPlacesLabel)
+        addSubview(line)
 
         let constraints = [ numberOfPlacesLabel.leadingAnchor.constraint(equalTo: self.leadingAnchor, constant: 20),
-                            numberOfPlacesLabel.bottomAnchor.constraint(equalTo: self.bottomAnchor, constant: -10)]
+                            numberOfPlacesLabel.bottomAnchor.constraint(equalTo: self.bottomAnchor, constant: -10),
+                            line.bottomAnchor.constraint(equalTo: self.bottomAnchor),
+                            line.leadingAnchor.constraint(equalTo: self.leadingAnchor),
+                            line.trailingAnchor.constraint(equalTo: self.trailingAnchor),
+                            line.heightAnchor.constraint(equalToConstant: 1.5)]
```
mcomella: If we have to specify an explicit height anchor, it could make HorizontalLineView confusing to use in the future.

Add a comment to the class?

PR #399, opened 2016-12-01
Closes #217: Add placeholder images when missing from image carousel

### LINKED ISSUE #217: if image is not present do not display carousel

mcomella:
#217


Prox/Prox/MapPlacesTransition.swift
```diff
@@ -161,7 +165,7 @@ class MapPlacesTransition: NSObject, UIViewControllerAnimatedTransitioning {
             fakeImageView.contentMode = UIViewContentMode.scaleAspectFill
             fakeImageView.clipsToBounds = true
             fakeImageView.backgroundColor = .red
```
sleroux: This might be the red that you're seeing. I must have left this in from debugging :(

PR #407, opened 2016-12-02
Issue 371 - Added 16px peek for the left card on the map carousel

sleroux:


Prox/Prox/PlaceCarousel/PlaceCarousel.swift
```diff
@@ -181,13 +181,14 @@ fileprivate class CarouselFlowLayout: UICollectionViewFlowLayout {
         }
         let pannedLessThanAPage = fabs(1 + currentPage - rawPageValue) > 0.5
         let flicked = fabs(velocity.x) > flickVelocity
+        let actualNextPage = pannedLessThanAPage && flicked ? nextPage : round(rawPageValue)
 
-        if pannedLessThanAPage && flicked {
-            pageDelegate?.didPage(toIndex: Int(max(0, nextPage)))
-            return CGPoint(x: nextPage * pageWidth, y: proposedContentOffset.y)
+        pageDelegate?.didPage(toIndex: Int(max(0, actualNextPage)))
```
mcomella: note: logic is slightly changed in that if `round(rawPageValue)` is < 0, it will return 0. This is probably what we want because we're using page indices.

PR #411, opened 2016-12-02
Fuffyemily/no bug event cache fixes

fluffyemily:
Fix event notification cache
Made logging quieter around remote config key fetching
added logging around event validation and notification sending


Prox/Prox/Models/Event.swift
```diff
@@ -172,12 +172,22 @@ class Event {
 
     internal func shouldShowEvent(withStartTime startTime: Date, endTime: Date?, timeIntervalBeforeStartOfEvent startTimeInterval: TimeInterval, timeIntervalBeforeEndOfEvent endTimeInterval: TimeInterval, atCurrentTime currentTime: Date) -> Bool {
         if isFutureEvent(currentTime: currentTime) {
-            return (startTime - startTimeInterval) <= currentTime
+            let validFutureEvent = (startTime - startTimeInterval) <= currentTime
+            if !validFutureEvent {
+                NSLog("\(description) is a future event but (\(startTime) - \(startTimeInterval)) <= \(currentTime)")
```
mcomella: nit: I'd prefer not to duplicate logic in logs – the logic can change slightly and the comment could become misleading! I'd prefer to describe intent, which changes less often than logic and list the params, e.g.:

`NSLog("\(description) is a future event but is not valid. startTime: \(startTime); startTimeInterval: \(startTimeinterval); currentTime: \(currentTime)")`

Here and below.

Prox/Prox/Utilities/EventNotificationsManager.swift
```diff
@@ -95,17 +95,17 @@ class EventNotificationsManager {
...
                 // check to see if there is an associated place
                 // don't send the notification if there isn't
                 let placesDB = PlacesProvider()
                 placesDB.place(forKey: event.placeId) { place in
-                    guard let _ = place else { return }
+                    guard let _ = place else { return  NSLog("Not sending notification for \(event.description) as it has no place") }
                     // check that travel times are within current location limits before deciding whether to send notification
                     TravelTimesProvider.canTravelFrom(fromLocation: currentLocation, toLocation: event.coordinates, before: event.arrivalByTime()) { canTravel in
-                        guard canTravel else { return }
+                        guard canTravel else { return NSLog("Not sending notification for \(event.description) as used cannot travel to it in time") }
```
mcomella: nit: -> user

Prox/Prox/Utilities/EventNotificationsManager.swift
```diff
@@ -102,10 +102,10 @@ class EventNotificationsManager {
                 // don't send the notification if there isn't
                 let placesDB = PlacesProvider()
                 placesDB.place(forKey: event.placeId) { place in
-                    guard let _ = place else { return  NSLog("Not sending notification for \(event.description) as it has no place") }
+                    guard let _ = place else { return  NSLog("Not sending notification for \(event.id) as it has no place") }
```
mcomella: fwiw, liuche is changing these to md5 hex hashes so it'll be hard to debug without querying Firebase directly – perhaps we should get a team python REPL going for Firebase, or something (but someone must have done that already!)?

mcomella:
One more nit: some of these logs are somewhat generic, e.g. "Not sending notifications for events because we have no location" – perhaps you want to log the method you're sending it from so it's never confused with another log statement?

PR #417, opened 2016-12-03
Closes #392: Display error message if we can't find location.

### LINKED ISSUE #392: Prox stalls on loading screen (missing location?)
Feedback from mheubusch@mozilla.com : Prox regular can't find any results near me. It had been able to in previous builds 

<table><tr><td>Created</td><td>Thu Dec 01 2016 20:05:45 GMT+0000 (UTC)</td></tr><tr><td>CFBundleShortVersion</td><td>1.0</td></tr><tr><td>CFBundleVersion</td><td>1</td></tr><tr><td>App uptime</td><td>271.5659</td></tr><tr><td>Build</td><td>584</td></tr><tr><td>Device type</td><td>iPhone 7</td></tr><tr><td>Device name</td><td>Michelle's iPhone</td></tr><tr><td>Screen size</td><td>375</td></tr><tr><td>Screen size</td><td>375px by 667px</td></tr><tr><td>Battery</td><td>34% Unplugged</td></tr><tr><td>Memory free</td><td>51 MB / 1532 MB</td></tr><tr><td>Disk free</td><td>102532 MB / 116315 MB</td></tr><tr><td>Network IP</td><td>192.168.0.14</td></tr></table>

 [Link to buddybuild feedback from build 584](https://dashboard.buddybuild.com/apps/57f51f8bcaa70a01005412ee/feedback?fid=58408299035c250100501b9e&bnum=584)

mcomella:
If the error is showing and we receive a location event, we will dismiss the
dialog and load the places.

Note that we check if a user has location services enabled already so this case
is if the user can't get a location but has services turned on (e.g. in a
basement, something horribly wrong).

#392.


Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -442,7 +458,17 @@ extension PlaceCarouselViewController: PlaceCarouselDelegate {
...
     func locationMonitor(_ locationMonitor: LocationMonitor, didUpdateLocation location: CLLocation) {
-        updateLocation(location: location)
+        if !isNoLocationAlertPresented {
+            updateLocation(location: location)
+            return
+        }
+
+        // If we don't dismiss the location alert, the detail view controller (called from
+        // updateLocation) does not get presented.
+        dismiss(animated: true) {
```
sleroux: Do you want to only call dismiss if we have the alert present?

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -457,11 +483,22 @@ extension PlaceCarouselViewController: LocationMonitorDelegate {
...
+        presentNoLocationAlert()
+    }
 }
 
 extension PlaceCarouselViewController: PlacesProviderDelegate {
     fileprivate func showErrorMessageIfNoPlaces() {
         if self.places.isEmpty {
+            let hasLocation = locationMonitor.getCurrentLocation() != nil
+            guard hasLocation else {
+                presentNoLocationAlert()
```
sleroux: Since this method is fired after a timeout fetching places, is it possible for this to be called when an alert is already being shown because of an error? If so, we might want to check to see if we are already showing one before showing another.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
mcomella: I was afraid of dismissing the PlaceDetailViewController

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
mcomella: afaik, the only error is shown below this but it's possible we'll add one.

Is it simple to check for another alert?

sleroux: You might be able to use the `presentedViewController` property of this view controller to see if it's the alert view you're showing. You'll probably have to have a reference to it to do the comparison though. Worst case, you can just use a boolean I suppose.

PR #418, opened 2016-12-03
Closes #324: Card view error state copy

### LINKED ISSUE #324: Updated copy for Card View Error State
screen shot 2016-11-28 at 1 13 46 pm

We'll use the same styles as the second line (regular, 14, 50% black) for the entire string.

mcomella:
#324.


mcomella:
Follow-up for error states distinction (as seen in commit summary) is #420.

PR #422, opened 2016-12-03
Add more analytics.

liuche:
This adds events for the "Read more" links, as well as some more details for event notifications. This does not include keeping a count of the number of event cards with banners - I was getting confused by all the paths, but I'll add it in another commit if I end up having time.


Prox/Prox/AppDelegate.swift
```diff
@@ -159,7 +159,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func application(_ application: UIApplication, didReceive notification: UILocalNotification) {
-        Analytics.logEvent(event: AnalyticsEvent.EVENT_NOTIFICATION, params: [:])
+        Analytics.logEvent(event: AnalyticsEvent.EVENT_NOTIFICATION, params: [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.FOREGROUND])
```
fluffyemily: This should be moved further down. Below here is a check to see if we are in the background (line 165) and if we are we do one thing, otherwise we do another. This function is called if the user is running iOS9. The others are called when the user is running iOS10.

`userNotificationCenter(_ center: ,willPresent notification:,withCompletionHandler:)` is called on IOS10 when the app is open (in the foreground) and `userNotificationCenter(center:,didReceive:,withCompletionHandler:)` is called on iOS10 when the user has responded to a notification when the app is in the background.

liuche:
Ah, sorry about that @fluffyemily. I made the changes you mentioned, and also added the background probe for iOS 9.

PR #439, opened 2016-12-07
No issue: set a max distance for the restaurants we show

mcomella:
I'll need to double-check this code over when I have a clear head (tomorrow). Given the risk, we could also write more tests.

**Note:** we should wait to merge this to master until we're sure we want it to make distribution easier.


fluffyemily:
This is one of those moments when I wish that we had gotten your fork running on buddybuild. We could have distributed this version to a couple of us to test in the wild for a few days.

How tricky would it be to cherry-pick these commits onto a branch on the master repo?

mcomella:
> How tricky would it be to cherry-pick these commits onto a branch on the master repo?

I pushed to the master repo: #444.

mcomella:
This is #444 now.

PR #442, opened 2016-12-07
Issue 440 - don't force unwrap index for place

fluffyemily:
We were force unwrapping result to optional function for index for place. I missed this during review


Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -629,7 +629,7 @@ class PlaceDetailViewController: UIViewController {
                 self.placeDetailsCardView(cardView: self.currentCardViewController.cardView, heightDidChange: self.currentCardViewController.cardView.frame.height)
 
                 self.currentCardViewController.beginAutoMovingOfCarousel()
-                let cardPos = (self.dataSource != nil) ? self.dataSource!.index(forPlace: self.currentCardViewController.place)! : -1
+                let cardPos = self.dataSource?.index(forPlace: self.currentCardViewController.place) ?? -1
```
mcomella: Change this in `pageToNextPlaceCard` too.

PR #445, opened 2016-12-07
Filter out places that are closed

mcomella:
Do we want to add a buffer? e.g. show places that will be open within an hour? I think it'd be fairly easy (just change `Date()` to add an hour).

Working on adding a quick test...


mcomella:
Not actively looking for this change, closing.

PR #448, opened 2016-12-07
No issue: Increase place radius to 1.5.

mcomella:
Land if Anthony agrees & also update FB.


mcomella:
Spoke with @liuche IRL – says it looks good.

PR #460, opened 2016-12-09
Add more event analytics

liuche:
This makes the event analytics more clear, and also covers some cases that had been missing.


liuche:
@fluffyemily I added my event analytics on top of your branch, so whatever buddybuild comes off of this build will have some useful analytics too (although on second thought, it will be mixed in with the analytics of the Prox release builds).

Prox/Prox/Utilities/AnalyticsUtilities.swift
```diff
@@ -61,11 +61,16 @@ public struct AnalyticsEvent {
     static let PARAM_ACTION       = "action"
     static let NUM_CARDS          = "num_cards"
     static let CARD_INDEX         = "card_index"
+    static let FOREGROUND         = "foreground"
```
mcomella: The value changed between this version and the previous, which could affect the existing data. Is this okay?

Prox/Prox/Utilities/EventNotificationsManager.swift
```diff
@@ -151,12 +151,15 @@ class EventNotificationsManager {
                     center.add(request) { error in
                         if let theError = error {
                             NSLog(theError.localizedDescription)
+                            Analytics.logEvent(event: AnalyticsEvent.ERROR_STATE, params: [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.EVENT_NOTIFICATION])
```
mcomella: It'd be great to capture the exact error that happens here (like how we capture the permissions error). Not sure it's worth doing now, but perhaps for future reference.

Prox/Prox/AppDelegate.swift
```diff
@@ -163,7 +163,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             let placeKey = notification.userInfo?[notificationEventPlaceIDKey] as? String {
             if ( application.applicationState == .inactive || application.applicationState == .background  ) {
                 placeCarouselViewController?.openPlace(placeKey: placeKey, forEventWithKey: eventKey)
-                Analytics.logEvent(event: AnalyticsEvent.EVENT_NOTIFICATION, params: [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.BACKGROUND])
+                Analytics.logEvent(event: AnalyticsEvent.EVENT_CLICKED, params: [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.BACKGROUND])
```
mcomella: Talking with @fluffyemily atm but [this SO post](http://stackoverflow.com/q/5099483) seems to indicate there are two entry points to the user from a user pressing a notification:

1. App is not running (via `didFinishLaunchingWithOptions`)
1. App is running in the BG (via `didReceive`)

It's possible `didReceive` is called in both cases.

fwiw, we don't have code for the former case to open the specific place so we should probably land this and add analytics when we fix it.

Prox/Prox/AppDelegate.swift
```diff
@@ -199,7 +199,7 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
         if response.notification.request.content.categoryIdentifier == "EVENTS" {
             if let eventKey = response.notification.request.content.userInfo[notificationEventIDKey] as? String,
                 let placeKey = response.notification.request.content.userInfo[notificationEventPlaceIDKey] as? String {
-                Analytics.logEvent(event: AnalyticsEvent.EVENT_NOTIFICATION, params: [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.CLICKED])
+                Analytics.logEvent(event: AnalyticsEvent.EVENT_CLICKED, params: [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.BACKGROUND])
```
mcomella: nit: should probably never happen but if the data in the notification, we won't track the analytics event even though the user interacted with it. We could prevent this by moving the analytics outside the if.

Prox/Prox/PlaceDetails/PlaceDetailsCardViewController.swift
```diff
@@ -223,7 +223,7 @@ class PlaceDetailsCardViewController: UIViewController {
         if !OpenInHelper.open(url: url) {
             print("lol unable to open web address")
         } else {
-            Analytics.logEvent(event: AnalyticsEvent.EVENT_BANNER_LINK, params: [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.CLICKED])
+            Analytics.logEvent(event: AnalyticsEvent.EVENT_BANNER_LINK, params: [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.EVENT_CLICKED])
```
mcomella: It's strange that `EVENT_CLICKED` is an event in the previous use case but `EVENT_CLICKED` here is a param.

Prox/Prox/Data/TravelTimesProvider.swift
```diff
@@ -83,7 +83,7 @@ struct TravelTimesProvider {
     }
 
     static func canTravelFrom(fromLocation: CLLocationCoordinate2D, toLocation: CLLocationCoordinate2D, before: Date, withCompletion completion: @escaping (Bool) -> ()) {
-        let timeInterval = before.timeIntervalSince(Date())
+        let timeInterval = Date().timeIntervalSince(before)
```
mcomella: I don't think this is correct. From the docs:

> The interval between the receiver and the another parameter. If the receiver is earlier than anotherDate, the return value is negative. If anotherDate is nil, the results are undefined.

`before` is expected to be a Date in the future: (`canTravelFrom(fromLocation:toLocation:before:withCompletion`), which means in the new code, the return value will be negative.

mcomella:
@fluffyemily Actually, I think the travel time code was correct before – see my comment above.

Prox/Prox/Utilities/AnalyticsUtilities.swift
liuche: Yes! I updated the new strings to match, so this should be okay. We never got any of this data anyway, which is why I made a bunch of changes.

Prox/Prox/AppDelegate.swift
liuche: +1 for landing this analytics and then adding more when we fix it.

Prox/Prox/AppDelegate.swift
liuche: I originally felt like we might not want to log if nothing happens, but that's also a good point.

Prox/Prox/Data/TravelTimesProvider.swift
liuche: I'm going to strip out this change bc I'm not sure if we'll be rebasing off these changes.

PR #476, opened 2016-12-23
Closes #365: Enable firebase persistence.

### LINKED ISSUE #365: Verify Firebase caches query results (for retryQueryPlaces)
Our `retryQueryPlaces` code assumes we have a cache – it queries the DB multiple times until the number of results stabilizes. If we don't cache the results between queries, that could be *a lot* of data use. We've turned FB persistence off (#321) so it's possible Firebase no longer caches.

Verify the Firebase queries are cached even with persistence off so we're not using too much data.

If the queries are not cached, consider re-enabling persistence instead of adding new, potentially buggy code.

Please do today or tomorrow so we're not adding too much code before the work week.

mcomella:
#365. This got a bit messy – let me know if there needs to be more explanation.

I would say this is messy enough that I'm not sure we should land on our current users, but given that the data is unlikely to change, we could just land it.

### DIFF
--- a/Prox/Prox.xcodeproj/project.pbxproj
+++ b/Prox/Prox.xcodeproj/project.pbxproj
@@ -81,6 +81,7 @@
 		E61831991DE0BDBA00D1810F /* TravelTimesView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E61831981DE0BDBA00D1810F /* TravelTimesView.swift */; };
 		E618319B1DE0C57700D1810F /* PlaceDetailsTravelTimesView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E618319A1DE0C57700D1810F /* PlaceDetailsTravelTimesView.swift */; };
 		E61972141DD28B7A00C9D093 /* PlaceDetailsEventView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E61972131DD28B7A00C9D093 /* PlaceDetailsEventView.swift */; };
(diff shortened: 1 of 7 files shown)


jhugman (changes requested):
1. I'm confused why you haven't done away with the `retryQueryPlaces` mechanism in its entirety, in favor of much simpler `query.observe`s.

The app requires a more or less continuous observer query on `venus/locations`, which then should trigger one-off downloads of the `venue/details/$id` documents.

To fit into the request/response theory of querying, we're unregistering observers to the `circleQuery` at [`FirebasePlacesDatabase`, L60][1]. This is exactly the query to be observing throughout the lifecycle of the foregrounded app. Getting rid of this polling behaviour so we're not fighting Firebase but working with it is the way forward here.

2. AFAICT, the way you're using `observeSingleEventButDownloadUpdates` means you're only getting the benefit of an updated cache when individual venues details are _changed_, but at the cost of many continuing observers (I don't know if there is a battery/network cost, I assume/hope not). 

The venue details are infrequently _changed_, but they are _added_, which you aren't going to observe if you're not observing the `circleQuery`.

This leaves me wondering if anything over and above [the line][2] 

```swift
db.persistenceEnabled = true
```
is actually doing anything, and why this didn't work satisfactorily in the first instance.

Changes requested:

1. work to remove [`circleQuery.removeAllObservers()`][1] 
2. make `circleQuery` a less transient object (i.e. keep hold of it, don't use `observeReady`.

I think the whole thing will drop out.

You might also want to move this work to a different PR, and land the [smallest patch][2] to minimally fulfill this issue.

 [1]: https://github.com/mozilla-mobile/prox/pull/476/files#diff-f0d396d25af9e0afac5ad995b523fdaeR60
 [2]: https://github.com/mozilla-mobile/prox/pull/476/files#diff-f13f708edc6b4f830829e407aba94f3cR95

PR #478, opened 2017-01-04
Move PlacesDataSource impl to PlacesProvider

fluffyemily:
In Hawaii I wondered whether we could make fetching faster and less data intensive if the `PlacesProvider` was also the `PlacesDataSource` that the rest of the app used. We could then check our geofire query results and only fetch those places that we were currently missing rather than fetching all places every time. This would also mean that if no new places were required then no fetches would be passed.

It's a thought experiment, but could be combined with the caching to provide an improvement to our place loading times.

### DIFF
--- a/Prox/Prox/Data/PlacesController.swift
+++ b/Prox/Prox/Data/PlacesController.swift
@@ -44,6 +44,8 @@ class PlacesProvider {
 
     fileprivate var firstFetch = true
 
+    fileprivate let placesLock = NSLock()
+
     func place(forKey key: String, callback: @escaping (Place?) -> ()) {
         database.getPlace(forKey: key).upon { callback($0.successResult() )}
     }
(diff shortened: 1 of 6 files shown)


Prox/Prox/Data/PlacesController.swift
```diff
@@ -246,6 +298,10 @@ extension PlacesProvider: PlaceDataSource {
     }
 
     func index(forPlace place: Place) -> Int? {
+        self.placesLock.lock()
```
mcomella: We need locks in `fetchPlace` (below this) too.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -168,14 +201,17 @@ class PlacesProvider {
     private func displayPlaces(places: [Place], forLocation location: CLLocation) {
         let filteredPlaces = PlaceUtilities.filterPlacesForCarousel(places)
         return PlaceUtilities.sort(places: filteredPlaces, byTravelTimeFromLocation: location, ascending: true, completion: { sortedPlaces in
+            self.placesLock.lock()
             self.places = sortedPlaces
             DispatchQueue.main.async {
                 self.delegate?.placesProvider(self, didReceivePlaces: self.places)
+                self.placesLock.unlock()
```
mcomella: From the docs:

> The NSLock class uses POSIX threads to implement its locking behavior. When sending an unlock message to an NSLock object, you must be sure that message is sent from the same thread that sent the initial lock message. Unlocking a lock from a different thread can result in undefined behavior.

Here, we're likely unlocking from a different thread.

Hack fix: Since we never directly modify the places array (but rather reassign the reference), lock, make a copy of the reference, unlock, and pass the copy to `didReceivePlaces` instead.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -44,6 +44,8 @@ class PlacesProvider {
 
     fileprivate var firstFetch = true
 
+    fileprivate let placesLock = NSLock()
```
mcomella: Move this declaration next to `var places` and add a comment to make it clear it should be used for all `places` accesses.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -515,24 +450,25 @@ extension PlaceCarouselViewController: PlacesProviderDelegate {
...
+        (self.presentedViewController as? PlaceDetailViewController)?.placesUpdated()
     }
 
     func placesProviderDidTimeout(_ controller: PlacesProvider) {
         showErrorMessageIfNoPlaces()
     }
+
+    func placesProviderDidFetchFirstPlace() {
+        // Wrap the openClosedPlace in an async block to make sure its queued after the
+        // carousel's refresh so the cells load before we invoke the transition
```
fluffyemily: I think the comment was originally added by @sleroux when he did the animations - I just copied it along with the relevant code when I moved it.

PR #479, opened 2017-01-04
WIP Issue 443 - Replace MKDirections with Google Distance Matrix API

fluffyemily:
MKDirections demonstrated itself to be pretty crap during Hawaii. I started an experiment to replace it with the Google Distance Matrix API to see if it would be any better. I have left the MKDirections code in to make it easy to swap between them for comparison purposes.

There will almost certainly have to be some tidying up done here. Also, the API is rate limited as described the [usage limits](https://developers.google.com/maps/documentation/distance-matrix/usage-limits), so we would have to think about paying for the service if we were to take this implementation forward.


Prox/Prox/Utilities/GoogleDirectionsMatrixTravelTimesProvider.swift
```diff
@@ -43,8 +46,71 @@ class GoogleDirectionsMatrixTravelTimesProvider: TravelTimesProvider {
...
+                            travelTime = TravelTimes(origin: fromLocation, destination: latLon, destinationPlaceKey: key, walkingTime: nil, drivingTime: value, publicTransportTime: nil)
+                        case MKDirectionsTransportType.transit:
+                            travelTime = TravelTimes(origin: fromLocation, destination: latLon, destinationPlaceKey: key, walkingTime: nil, drivingTime: nil, publicTransportTime: value)
+                        default:
+                            travelTime = TravelTimes(origin: fromLocation, destination: latLon, destinationPlaceKey: key, walkingTime: value, drivingTime: nil, publicTransportTime: nil)
+                        }
+                        allTravelTimes.append(travelTime)
+                    }
+                    if expectedResponseCount == 0 {
+                        deferred.fill(with: DatabaseResult.succeed(value: allTravelTimes))
```
mcomella: I'm not sure which thread data task runs on (the session manager's but assuming it's concurrent, I'm concerned this can be called early. We increment `expectedResponseCount`, decrement it (in the data task response), and check `== 0` before the while loop iterates again.

It looks like we can calculate expectedResponseCount in advance – perhaps we should do that? (unless you know the response handler is called on the same thread

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -23,31 +23,40 @@ struct PlaceUtilities {
...
                     return placeAETA <= placeBETA
                 } else {
                     return placeAETA >= placeBETA
                 }
             }
-
-            for (index, place) in sortedByTravelTime.enumerated() {
-                sortedPlaces[index] = place
+            let sortedByTravelTime: [Place] = sortedTravelTimes.flatMap { time in
+                let placeIndex = sortedByDistance.index { place in
```
mcomella: As in the other PR, I think `Array.index` could be linear search and since we're iterating over the places, this could be n<sup>2</sup> run time.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -23,31 +23,40 @@ struct PlaceUtilities {
...
+                    return time.destinationPlaceKey == place.id
+                        || (time.destination?.latitude == place.latLong.latitude && time.destination?.longitude == place.latLong.longitude)
```
mcomella: What are we checking here (with the coordinates)? Can you add a comment or store in a named temporary variable to clarify?

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -23,31 +23,40 @@ struct PlaceUtilities {
...
                 if ascending {
                     return placeAETA <= placeBETA
                 } else {
                     return placeAETA >= placeBETA
                 }
             }
-
-            for (index, place) in sortedByTravelTime.enumerated() {
-                sortedPlaces[index] = place
+            let sortedByTravelTime: [Place] = sortedTravelTimes.flatMap { time in
```
mcomella: nit: I think there's stronger intent when using `filter` to remove items from a collection than `flatMap`.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -24,13 +24,15 @@ struct PlaceUtilities {
...
         var placeLocations = [PlaceKey: CLLocationCoordinate2D]()
-        sortedByDistance.forEach { placeLocations[$0.id] = $0.latLong }
+        placesWithoutTimes.forEach { placeLocations[$0.id] = $0.latLong }
 
+        // this means we will probably only get walking directions for places, but I think that will be OK for now
         travelTimesProvider.travelTimes(fromLocation: location.coordinate, toLocations: placeLocations, byTransitType: .walking).upon { result in
-            guard let travelTimes = result.successResult(),
-            !travelTimes.isEmpty else { return completion(sortedByDistance) }
+            guard var travelTimes = result.successResult() else { return completion(sortedByDistance) }
+            travelTimes += placesWithTimes.flatMap { $0.cachedTravelTime(fromLocation: location) }
```
mcomella: nit: I think `flatMap` is redundant – the `placesWithTimes` collection is created with the flatMap function.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -48,6 +50,7 @@ struct PlaceUtilities {
                 }
                 if placeIndex != nil {
                     let place = sortedByDistance[placeIndex!]
+                    Place.travelTimesCache[place.id] = CachedTravelTime(Deferred(filledWith: DatabaseResult.succeed(value: time)), location)
```
mcomella: `flatMap` comes from a functional background so it's a little unexpected for it to have side effects but if this is what is simplest, so be it.

Prox/Prox/Utilities/TravelTimesProvider.swift
```diff
@@ -0,0 +1,40 @@
...
+import CoreLocation
+
+var travelTimePadding: Double = {
+    return RemoteConfigKeys.travelTimePaddingMins.value * 60
+}()
+
+var MIN_WALKING_TIME: Int = {
+    // Note that this is semantically maximum walking time,
+    // rather than minimum walking time (as used throughout the codebase).
+    return RemoteConfigKeys.maxWalkingTimeInMins.value
```
fluffyemily: I didn't add this comment, only copied it from inside `MKDirectionsTravelTimesProvider`

Prox/Prox/Utilities/TravelTimesProvider.swift
```diff
@@ -0,0 +1,40 @@
...
+}()
+
+var YOU_ARE_HERE_WALKING_TIME: Int = {
```
fluffyemily: again, I think this is legacy copied over from the original implementation but I shall address

Prox/Prox/Utilities/GoogleDirectionsMatrixTravelTimesProvider.swift
```diff
@@ -43,8 +46,71 @@ class GoogleDirectionsMatrixTravelTimesProvider: TravelTimesProvider {
         return AFURLSessionManager(sessionConfiguration: configuration)
     }()
 
+    static func travelTimes(fromLocation: CLLocationCoordinate2D, toLocations: [PlaceKey : CLLocationCoordinate2D], byTransitType transitType: MKDirectionsTransportType) -> Deferred<DatabaseResult<[TravelTimes]>> {
+        NSLog("Calculating travel times for \(toLocations)")
+        let locations = Array(toLocations.values)
+        var index = 0
+        var fetchSize = min(locations.endIndex, 25) - 1
```
fluffyemily: I'll add a comment, but it's because the API has a limit on 25 destinations per request.

Prox/Prox/Utilities/TravelTimesProvider.swift
fluffyemily: Globals are lazy by default. I should move these to `AppConstants` though

Prox/Prox/Utilities/PlaceUtilities.swift
fluffyemily: But we are also mapping `TravelTimes` to `Place`'s as well as filtering out times that we cannot map to places so flatmap prevents us from having to filter and then map which would increase the number of interations

Prox/Prox/Utilities/PlaceUtilities.swift
fluffyemily: We cannot guarentee that every `TravelTime` returned will have a place id (it depends on which function we used to fetch the `TravelTime`. Therefore if the `TravelTime` has no place id, we have to use the `TravelTime` destination coordinates to match times with places instead.

PR #482, opened 2017-01-10
Closes #481: Disable notifications

### LINKED ISSUE #481: Disable notifications for Chicago test
Thinking maybe we should have a Chicago/Nashville branch and land this on that but perhaps we're not doing any other breaking changes like this and it's too much overhead.

mcomella:
#481.


Prox/Prox/AppDelegate.swift
```diff
@@ -149,26 +141,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func applicationWillTerminate(_ application: UIApplication) {
         // Called when the application is about to terminate. Save data if appropriate. See also applicationDidEnterBackground:.
         placeCarouselViewController?.locationMonitor.cancelTimeAtLocationTimer()
-        eventsNotificationsManager.persistNotificationCache()
         AppState.exiting()
     }
 
     func application(_ application: UIApplication, performFetchWithCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
```
fluffyemily: If we're going this far, this entire function will never get called with `application.setMinimumBackgroundFetchInterval(AppConstants.backgroundFetchInterval)` removed so you may as well remove the entire thing

mcomella:
I think I mistakenly took, "Don't comment it out, it's in version control!" too far because you're right – it makes it harder to re-add (once conflicts arise). That being said, I like the feature flag idea – I'll look into it.

mcomella:

I realized, if we have add tests that require 'areNotificationsEnabled == true`, we may need to set the feature flag based on whether or not we're testing - but we can handle that when we get to that.

CC @liuche @thebnich for awareness that I'm throwing in a feature flag that could affect testing.

PR #484, opened 2017-01-11
No issue: Create Prox Chicago build to help us 

mcomella:
It adds a table view controller so users can select which location to fake.

I don't like that branches initing the view controller in the AppDelegate depending on whether or not we're faking location but it was the simplest thing to do.

Includes work from the rm ty dialog branch (github :( ).

Screenshot of fake locations:
simulator screen shot jan 10 2017 17 49 30


mcomella:
Closing in lieu of #504.

PR #485, opened 2017-01-14
Remove unused sessions, correct loading and permissions

liuche:
This fixes the biggest problem with session measuring where we were collecting data even if the app isn't visible. This shouldn't be a problem (while notifications are disabled) but this fixes differentiating between background sessions (testing with the Xcode "background fetch" scheme) and user-initiated sessions.

This also adds measurements for permissions, and also moved the analytics collection into user-initiated actions rather than including background fetches.

(I initially thought that this would require setting listeners for viewDidLoad but I had misconceptions about how granular iOS view management was, so that ended up not being relevant.)

### DIFF
--- a/Prox/Prox/Models/AppState.swift
+++ b/Prox/Prox/Models/AppState.swift
@@ -23,20 +23,6 @@ class AppState {
         updateSessionState(newState: State.loading)
     }
 
-    static func requestPermissions() {
-        if (state == State.loading) {
-            // Stop loading session if handling permissions
-            let params: [String: Any] = [AnalyticsEvent.PARAM_ACTION: AnalyticsEvent.PERMISSIONS]
-            Analytics.endSession(sessionName: State.loading.rawValue + AnalyticsEvent.SESSION_SUFFIX, params: params)
-            state = State.permissions
-            print("[debug] endsession because permissions. New state: \(state)")
-        } else {
-            let params : [String: Any] = [AnalyticsEvent.SESSION_STATE: state.rawValue]
-            Analytics.logEvent(event: AnalyticsEvent.PERMISSIONS, params: params)
-            print("[debug] permissions event. Current state \(state)")
-        }
-    }
-
     static func enterBackground() {
         print("[debug] entering background from: " + state.rawValue)
         preBackgroundState = state
--- a/Prox/Prox/Models/AppState.swift
+++ b/Prox/Prox/Models/AppState.swift
@@ -6,17 +6,35 @@ import Foundation
 
 class AppState {
     enum State: String {
-        case loading, details, carousel, background, exiting, unknown
+        case initial, loading, permissions, details, carousel, background, unknown
     }
 
     private static var cardVisits = Set<Int>()
 
     // Initial app state
-    private static var state = State.loading
+    private static var state = State.initial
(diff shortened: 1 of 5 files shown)


Prox/Prox/Models/AppState.swift
```diff
@@ -6,19 +6,37 @@ import Foundation
...
     static func getState() -> State {
         return state
     }
 
+    static func enterLoading() {
+        updateSessionState(newState: State.loading)
+    }
+
+    static func requestPermissions() {
+        if (state == State.loading) {
```
mcomella: Why do we sometimes consider permissions a session and sometimes consider permissions an event? I'm concerned it'd be hard to accurately track permissions requests this way.

Also, if we're displaying a dialog, I think we'd still be loading under the hood so I'd think we should continue the loading session (fwiw, I don't think we do any cleanup in `ResignActive` or `EnterBackground` so I think we'd continue loading during our ~5s background execution time too) – why do we cancel the loading session here?

Prox/Prox/Utilities/AnalyticsUtilities.swift
```diff
@@ -17,8 +17,10 @@ class Analytics {
             print("No Flurry key! Not collecting analytics.")
             return
         }
+
+        // Deprecated, but no Swift alternative right now
```
liuche: Good call - using the Swift alternative now.

Prox/Prox/Utilities/AnalyticsUtilities.swift
```diff
@@ -17,8 +17,10 @@ class Analytics {
...
+        Flurry.setSessionContinueSeconds(30)
         Flurry.startSession(flurryKey)
```
liuche: I tested this "30 seconds" continuation parameter with the Swift call and verified that it still works on the Flurry live dashboard.

Prox/Prox/Models/AppState.swift
```diff
@@ -6,19 +6,37 @@ import Foundation
...
     }
 
     private static var cardVisits = Set<Int>()
 
     // Initial app state
-    private static var state = State.loading
+    private static var state = State.initial
     private static var preBackgroundState: State?
 
     static func getState() -> State {
```
liuche: This sounded a little complicated, so I passed on doing this right now, but I'll look into it next.

Prox/Prox/Models/AppState.swift
liuche: This was originally so that showing the permissions dialog didn't count as "loading" time, but I ended up replacing this somewhat complicated logic with events that fire when we show a dialog reprompting the user (if they've denied location permissions). This might throw off the loading time a bit for the first session (when the user has to decide which permissions to grant Prox), but I think that's okay.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -303,6 +302,7 @@ class PlaceCarouselViewController: UIViewController {
         alertController.addAction(settingsAction)
         alertController.addAction(quitAction)
 
+        AppState.requestPermissions()
```
liuche: Removed permissions sessions entirely in favor of recording dialog-based events.

PR #507, opened 2017-02-07
Closes #506 - Iterate through provider sources on the client

### LINKED ISSUE #506: Update client to use new Chicago data
In an effort to simplify incremental crawling, we're now removing all top-level fields from each venue and storing all provider-specific data under each venue's `providers` branch. This simplifies the crawling code, and the responsibility of choosing which providers are used for which fields is moved to the client.

thebnich:
A big diff, but most of this is straightforward refactoring that moves the Firebase data extraction from `Place.swift` into the new `PlaceProviders.swift`. This creates a new protocol, `PlaceProvider`, that contains all of the fields a provider might have. Note that all fields are optional, so there's no requirement that any provider gives any particular fields.

The primary implementation of this protocol is `SourcePlaceProvider`, which pulls provider info from the Firebase dict (most of this was simply taken from the `Place` initializer). The other implementation is the `CompositePlaceProvider`, which takes a list of providers (ordered by priority) and sets each field based on the first source where it's available. Since Yelp is currently the only provider for each place, the `CompositePlaceProvider` will just be a mirror of Yelp results, but this will of course change once TA/Wikipedia are added.

Once we have our `CompositePlaceProvider`, we can populate our `Place` fields as normal. Like before, we require that we have at least an ID, name, and lat/long pair for a place.

Note: I think Yelp data will divided into both Yelp v2 and Yelp v3 data in Firebase (under `providers/yelp` and `providers/yelpv3`), where we currently have only v2 data. I don't think we want separate Yelp v2/v3 providers exposed in a `Place`, though, so I expect that once we have v3 data, `yelpProvider` will itself be a `CompositePlaceProvider` populated from the `providers/yelp` and `providers/yelpv3` branches.


Prox/Prox/Models/PlaceProviders.swift
```diff
@@ -0,0 +1,159 @@
...
+                url = provider.url
+            }
+
+            if address == nil {
+                address = provider.address
+            }
+
+            if hours == nil {
+                hours = provider.hours
+            }
```
mcomella: Maybe comment that rating & totalReviewCount are intentionally not filled in?

Prox/Prox/Models/PlaceProviders.swift
```diff
@@ -0,0 +1,159 @@
...
+
+extension SourcePlaceProvider {
+    fileprivate static func getCategories(fromFirebaseValue value: [[String:String]]) -> (names: [String], ids: [String]) {
+        var names = [String]()
+        var ids = [String]()
+        for category in value {
+            guard let name = category["text"],
+                let id = category["id"] else {
+                    print("lol unable to retrieve category from firebase data for place")
+                    return ([], [])
```
mcomella: Maybe we should `continue` rather than `returning` early if we're missing just one name/category pair.

Prox/Prox/PlaceDetails/PlaceDetailsCardViewController.swift
```diff
@@ -246,7 +242,10 @@ class PlaceDetailsCardViewController: UIViewController {
     }
 
     @objc private func autoMoveToNextCarouselImage() {
-        self.imageCarouselCollectionView.scrollToItem(at: IndexPath(item: getNextCarouselPageIndex(), section: 0), at: UICollectionViewScrollPosition.centeredHorizontally, animated: true)
+        return
+
+        // TODO: Re-enable after we have photos from Yelp v3 and other sources.
```
mcomella: Add a GH issue; reference # here.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -57,10 +57,13 @@ struct PlaceUtilities {
     }
 
     static func filterPlacesForCarousel(_ places: [Place]) -> [Place] {
+        return places
+
+        // TODO: Re-enable category filtering once we have Yelp v3 and other data.
```
mcomella: Same: add an issue.

Prox/Prox/Models/PlaceProviders.swift
```diff
@@ -0,0 +1,159 @@
...
+    var latLong: CLLocationCoordinate2D? { get }
+    var photoURLs: [URL] { get }
+    var url: URL? { get }
+    var address: String? { get }
+    var hours: OpenHours? { get }
+    var rating: Float? { get }
+    var totalReviewCount: Int? { get }
+}
+
+class SourcePlaceProvider: PlaceProvider {
```
mcomella: nit: `SourcePlaceProvider` is a confusing name to me – I was thinking "it's the source of everything" (singular) rather than "here is one of our sources, like Yelp".

I think the implementation should really be called, `PlaceProvider` but the interface gets in the way. We could do an ugly: `IPlaceProvider` or `PlaceProviderProtocol` (I'm not a fan of `*Impl`). Or maybe the implementation could be called `SinglePlaceProvider` in contrast to `CompositePlaceProvider`?

Then again, xcode makes refactoring bullshit so maybe just leave it as is. :)

Prox/Prox.xcodeproj/project.pbxproj
```diff
@@ -353,11 +353,11 @@
 		7B5095631DB531CA007C54F0 /* Models */ = {
 			isa = PBXGroup;
 			children = (
-				E65857DD1DB69F4200CD18F8 /* ReviewProvider.swift */,
-				7B5095641DB531CA007C54F0 /* Place.swift */,
-				7B48EBD91DD1F679005C36CB /* GeofenceRegion.swift */,
-				7BA05BE41DD4CCB1009AC121 /* Event.swift */,
 				2380CA901DE4140A002D73F1 /* AppState.swift */,
+				7BA05BE41DD4CCB1009AC121 /* Event.swift */,
```
thebnich: Yeah, they might -- it happens all the time :) Generally, they're pretty easy to resolve manually.

Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -36,7 +36,7 @@ class RemoteConfigKeys {
 
     // The search radius the app will use to query the Firebase database.
     // This is measured in kilometers.
-    public static let searchRadiusInKm = RemoteConfigRadius(key: "search_radius_in_km", defaultValue: 1.5)
+    public static let searchRadiusInKm = RemoteConfigRadius(key: "search_radius_in_km", defaultValue: 0.2)
```
mcomella: Make sure to update this in Firebase too.

mcomella:
btw, if you name your commits "Closes #<whatever>" instead of "Issue #<whatever>", the associated github issue will close automatically when the changes are merged.

Prox/Prox/Models/PlaceProviders.swift
thebnich: `SinglePlaceProvider` sounds reasonable to me.

PR #512, opened 2017-02-07
No issue: Logging cleanup

thebnich:
Some simple changes to make our logging a bit less ad hoc. This allows us to
* use standard logging level for each message (debug/info/warn/error)
* filter levels based on the build
* not have to include the tag in each message
* create new `Logger` instances with different tags

I'm also not a fan of the current "lol" tag since it has no meaning, but we can fix that separately. This at least sets the foundation so we can change that default tag easily.

### DIFF
--- a/Prox/Prox/Models/PlaceProviders.swift
+++ b/Prox/Prox/Models/PlaceProviders.swift
@@ -152,7 +152,7 @@ extension SinglePlaceProvider {
         var ids = [String]()
         for category in value {
             guard let name = category["text"], let id = category["id"] else {
-                print("lol unable to retrieve category from firebase data for place")
+                log.warn("unable to retrieve category from firebase data for place")
                 continue
             }
 
--- a/Prox/Prox/Utilities/Logger.swift
+++ b/Prox/Prox/Utilities/Logger.swift
@@ -1,0 +1,66 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+enum LoggerLevel: Int {
+    case debug
+    case info
+    case warn
+    case error
+}
+
+class Logger {
+    private let tag: String
+
+    var minPrintedLevel: LoggerLevel
+
+    /// Create a Logger instance that prints messages with the given tag.
+    /// Messages are filtered out if they are less than the given level.
+    init(tag: String, minPrintedLevel: LoggerLevel) {
+        self.tag = tag
+        self.minPrintedLevel = minPrintedLevel
+    }
+
+    /// Create a Logger instance that prints messages with the given tag.
+    /// Assigns the level based on the current build channel.
+    convenience init(tag: String) {
+        let minPrintedLevel: LoggerLevel
+        switch AppConstants.BuildChannel {
+        case .Debug:
+            minPrintedLevel = .debug
+        case .Release:
+            minPrintedLevel = .warn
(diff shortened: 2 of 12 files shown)


Prox/Prox/Utilities/Logger.swift
```diff
@@ -0,0 +1,66 @@
...
+    case info
+    case warn
+    case error
+}
+
+class Logger {
+    private let tag: String
+
+    /// All log levels >= this level will be printed.
+    var level: LoggerLevel
```
mcomella: nit: maybe -> `minPrintedLevel` or `maxPrintedLevel` and do away with the comment (which is more likely to go out of date)?

Prox/Prox/Utilities/Logger.swift
```diff
@@ -0,0 +1,66 @@
...
+
+    /// Create a Logger instance that prints messages with the given tag.
```
mcomella: Is `///` official Swift documentation style?

Prox/Prox/Utilities/Logger.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// Assigns the level based on the current build channel.
+    convenience init(tag: String) {
+        let level: LoggerLevel
+        switch AppConstants.BuildChannel {
+        case .Debug:
+            level = .debug
+        case .Release:
+            level = .warn
+        default:
```
mcomella: If we add new build configurations, we'll probably forget to updated the logger. I'd rather enumerate all the BuildChannels so the compile fails rather than forget to set the log levels.

Prox/Prox/Utilities/Logger.swift
```diff
@@ -0,0 +1,66 @@
...
+
+    func warn(_ message: Any?) {
+        log(message, levelTag: "warn", level: .warn)
+    }
+
+    func error(_ message: Any?) {
+        log(message, levelTag: "error", level: .error)
+    }
+
+    private func log(_ message: Any?, levelTag: String, level: LoggerLevel) {
```
mcomella: nit: Shouldn't we convert the `LoggerLevel` to a string rather than use a separate string? Maybe you can add a String arg to each log level.

Prox/Prox/Utilities/Logger.swift
```diff
@@ -0,0 +1,66 @@
...
+        guard level.rawValue >= self.level.rawValue else { return }
+        print("\(tag) [\(levelTag)] \(message ?? "nil")")
```
mcomella: Should we use `NSLog` so these show up on BB? My understanding is that `NSLog` does but `print` does not.

Prox/Prox/Models/PlaceProviders.swift
```diff
@@ -145,9 +145,8 @@ extension SourcePlaceProvider {
         var ids = [String]()
         for category in value {
             guard let name = category["text"],
-                let id = category["id"] else {
-                    print("lol unable to retrieve category from firebase data for place")
```
mcomella: nit: did you intend to rm this print?

Prox/Prox/Utilities/Logger.swift
thebnich: Hm, I didn't realize stringifying `Logger.debug` automagically returns "debug", so I can just use that.

Prox/Prox/Models/PlaceProviders.swift
thebnich: I did because I was thinking that a place won't necessarily have a category, but looking at this again, this is only logged if a place _does_ have a category but the necessary fields are missing. I'll add it back.

PR #515, opened 2017-02-09
No issue: Generate LocalConfig.plist

thebnich:
The current behavior for `AppConstants.firebaseRoot` isn't great. For one, the default path is "", meaning we try to read from the top level of the Firebase tree. Secondly, changing this path to point somewhere else (e.g., my own `brian/` branch) gets caught in version control.

I'd like to propose a local git-ignored config file, `LocalConfig.plist`. We can generate this file as part of a post-clone script, created in this PR, which includes `pod install`.

Note that I'm using `whoami` to set the login user name as the branch like we do in `prox-server`. If it's more useful, I could be convinced to make this script generate a plist that points to the shared `development/` instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ xcuserdata/
 Prox/Prox/GoogleService-Info.plist
 Prox/Prox/APIKeys.plist
 
+# Local config
+Prox/Prox/LocalConfig.plist
+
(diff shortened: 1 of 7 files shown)


thebnich:
Not worth it the added complexity with the recent fixes to debug branches.

PR #517, opened 2017-02-10
No issue: update firebase root to new schema.

mcomella:


Prox/Prox/Utilities/AppConstants.swift
```diff
@@ -75,13 +75,10 @@ public struct AppConstants {
...
     public static let firebaseRoot: String = {
-        switch (BuildChannel) {
-        case .CurrentLocation, .MockLocation, .Release:
-            return "production/"
-
-        case .Debug:
-            return ""
-        }
+        let root = FirebaseBranches.N02_CHICAGO
+        //let root = FirebaseBranches.getBranch(forUser: "jane") // for debugging.
```
thebnich: Nit: remove

mcomella: I'd rather leave it – I doubt anyone will use the fn if I remove the statement, and then they're more likely to screw it up and waste time (e.g. forgetting the trailing slash, typoing "users").

PR #520, opened 2017-02-13
Closes #514 - Add Yelp v3 and re-enable photos/categories

### LINKED ISSUE #514: Re-enable photos after crawling Yelp v3
The current places base is built on Yelp v2, which doesn't include photos. We should turn photos back on once we have v3 data.

thebnich:
I _think_ this should work, but the Firebase DB hasn't been updated to include Yelp v3 yet, so all the places currently have empty categories. This should be ready for review once places start showing up.

### DIFF
--- a/Prox/Prox/Models/PlaceProviders.swift
+++ b/Prox/Prox/Models/PlaceProviders.swift
@@ -42,11 +42,14 @@ class SinglePlaceProvider: PlaceProvider {
             self.description = nil
         }
 
-        if let categoriesFromFirebase = dict["categories"] as? [[String:String]] {
-            categories = SinglePlaceProvider.getCategories(fromFirebaseValue: categoriesFromFirebase)
-        } else {
-            categories = ([], [])
+        let categoryIds = dict["categories"] as? [String] ?? []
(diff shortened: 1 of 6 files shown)


thebnich:
Update: seems to be working as expected.

mcomella:
> Update: seems to be working as expected.

😎

Prox/Prox/Models/PlaceProviders.swift
```diff
@@ -42,11 +42,14 @@ class SinglePlaceProvider: PlaceProvider {
             self.description = nil
         }
 
-        if let categoriesFromFirebase = dict["categories"] as? [[String:String]] {
-            categories = SinglePlaceProvider.getCategories(fromFirebaseValue: categoriesFromFirebase)
-        } else {
-            categories = ([], [])
+        let categoryIds = dict["categories"] as? [String] ?? []
+        var categories = (names: [String], ids: [String])([], [])
```
mcomella: super nit: no need to waste time changing now but consider for next time. I think this would be clearer:
```swift
var categories: (names: [String], ids: [String]) = ([], [])
```

PR #523, opened 2017-02-15
No issue: Add MapViewController with gmaps

mcomella:
If I land this, I expect it'll make it easier to work on simultaneously. Also, let's review early and (maybe not so) often!

TODO (for follow-up):
* Add places to map
* Have places in map update as we scroll
* (assuming it's in the design) Have places in carousel update to map frame of reference. (this item and the one above might conflict with the place carousel refactor)
* Once Brian works on removing the place carousel, I expect this to move into a UINavigationController with PlaceDetailsViewController at the base. That being said, this depends on the final designs
* @thebnich Will this conflict with your filtering changes? Since I'm using the map view button for, uh, the map view 😉


Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -699,6 +700,10 @@ class PlaceDetailViewController: UIViewController {
         return nextCardImageCarousel
     }
 
+    @objc fileprivate func openMapView() {
```
thebnich: Nit: `fileprivate` -> `private`

Prox/Prox/AppDelegate.swift
```diff
@@ -85,6 +87,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         FIRDatabase.database().persistenceEnabled = false
     }
 
+    private func setupGoogleMaps() {
```
thebnich: Grammar nit: `setup` -> `setUp`

Prox/Prox/Map View/MapViewController.swift
```diff
@@ -3,14 +3,65 @@
...
 
 import UIKit
+import GoogleMaps
 
 class MapViewController: UIViewController {
 
-    init() {
-        super.init(nibName: nil, bundle: nil)
-    }
+    lazy var rootContainer: UIStackView = {
```
thebnich: Nit: `private` on these `lazy var`s.

Prox/Prox/Map View/MapViewController.swift
```diff
@@ -3,14 +3,65 @@
...
+                rootContainer.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+                rootContainer.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+                rootContainer.bottomAnchor.constraint(equalTo: bottomLayoutGuide.topAnchor),
+        ]
+
+        NSLayoutConstraint.activate(constraints, translatesAutoresizingMaskIntoConstraints: false)
+    }
+
+    @objc private func close() {
+        print("called")
```
thebnich: Remove

Prox/Prox/Map View/MapViewController.swift
thebnich: Also, you can make these non-`lazy` (since they'll be created immediately anyway), which will allow you to make them immutable.

mcomella: The `lazy var` convention appears to be done so you can declare the variables/methods in the file in any order (whereas you have to wait for references to be defined first if it's non-lazy).

PR #525, opened 2017-02-15
Closes #508 - Implement place filters

### LINKED ISSUE #508: Add "V1" filtering UI
home - filter_final

A simple check-box type of functionality to start. This will allow us to layer on more functionality as we get more feedback too.

Assets to come

thebnich:
Adds filter support to the data source (`PlacesProvider`), and implements UI (`FilterViewController`) for adjusting the filters.


thebnich:
Since this PR was getting pretty big and it'd be good to land all these changes ASAP, I didn't finish the animation polish described in #508. Opened #526 for updating the animation.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -139,6 +129,28 @@ class PlacesProvider {
         }
     }
 
+    func filterPlaces(filters: [PlaceFilter]) -> [Place] {
+        let enabledCategories = Set(filters.filter { $0.enabled }.map { $0.categories }.reduce([], +))
+        let toRoots = CategoriesUtil.categoryToRootsMap
+
+        return allPlaces.filter { place in
```
mcomella: nit: should we lock for access to allPlaces?

Note: this is called from `updateDisplayedPlaces` which expects the caller to lock so this method would also need to be locked by the called (but it's public!).

Prox/Prox/Data/PlacesController.swift
```diff
@@ -257,14 +270,20 @@ extension PlacesProvider: PlaceDataSource {
...
     }
+
+    func refresh() {
+        assert(Thread.isMainThread)
+
+        placesLock.withWriteLock {
+            updateDisplayedPlaces()
+        }
+
+        delegate?.placesProvider(self, didUpdatePlaces: displayedPlaces)
```
mcomella: We take a read lock when we call this in `displayPlaces`. As I mention earlier, maybe we should just return a copy.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -30,32 +30,22 @@ class PlacesProvider {
...
-            for (index, place) in displayedPlaces.enumerated() {
-                placesMap[place.id] = index
-            }
-            self.placeKeyMap = placesMap
-        }
-    }
-    fileprivate var placeKeyMap = [String: Int]()
+    private var allPlaces = [Place]()
 
+    fileprivate var displayedPlaces = [Place]()
```
mcomella: I wonder if there would be a benefit separating the data (allPlaces) from the representation/transformations (displayedPlaces) in two separate classes. So we have a PlacesFetcher which does the fetch, then we have a PlacesPresenter which can sort, filter, etc. data from the PlacesFetcher. _Something something single responsibility principle..._ :P

Prox/Prox/PlaceDetails/FilterViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+        background.backgroundColor = Colors.filterOverlayDim
+        background.alpha = 0
+        background.addTarget(self, action: #selector(didPressClose), for: .touchDown)
+        view.addSubview(background)
+
+        let container = UIView()
+        container.backgroundColor = .white
+        view.addSubview(container)
+
+        placeCountLabel.font = UIFont.systemFont(ofSize: 20)
```
mcomella: nit: we have a Fonts file for defining these (but I'm not convinced it's particularly useful).

Prox/Prox/PlaceDetails/FilterViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+        delegate?.filterViewController(self, didUpdateFilters: filters)
+    }
+
+    @objc private func didPressClose() {
+        dismiss(animated: true, completion: nil)
+    }
+
+    override func dismiss(animated flag: Bool, completion: (() -> Void)? = nil) {
+        super.dismiss(animated: flag, completion: completion)
+        delegate?.filterViewController(self, didDismissWithFilters: filters)
```
mcomella: Neat override!

But what if we want to dismiss a UIAlert? Would this get called? If so, maybe we want to put this in `didPressClose`

Prox/Prox/PlaceDetails/FilterViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    fileprivate override var isSelected: Bool {
+        didSet {
+            super.isSelected = isSelected
+            if isSelected {
+                setTitleColor(Colors.filterButtonCheckedForeground, for: .normal)
+                setImage(#imageLiteral(resourceName: "icon_selected"), for: .normal)
```
mcomella: nit: does this reload the image from disk every time? If we see perf issues, we might want to consider caching it.

Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -66,10 +66,10 @@ class PlaceDetailViewController: UIViewController {
         return panGesture
     }()
     
-    lazy var mapButton: MapButton = {
-        let button = MapButton()
+    lazy var mapButton: UIButton = {
+        let button = UIButton()
+        button.setImage(#imageLiteral(resourceName: "button_map"), for: .normal)
```
thebnich: `#imageLiteral` shows up automagically in Xcode, with the image inlined directly in the source (you don't actually see this `#imageLiteral` if you look at the source in Xcode).

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -16,6 +16,16 @@ protocol PlaceDataSource: class {
...
     func sortPlaces(byLocation location: CLLocation)
+
+    /// Returns all nearby places filtered by the given set of filters.
+    /// This function does not modify the data source.
+    func filterPlaces(filters: [PlaceFilter]) -> [Place]
+
+    /// Refreshes the data source by filtering with the enabled filters.
+    func refresh()
+
+    /// You must call refresh() after changing the enabled state of any filter!
```
thebnich: Not quite sure what you're suggesting here. `PlacesProvider` has multiple filters, so what would your proposed `enabled` field look like? Would it take an array of the same cardinality as the filter array, with correlated indices?

Prox/Prox/Data/PlacesController.swift
thebnich: I agree, but didn't bother here since I noticed that `PlacesProvider` was already doing sorting -- and I figured this PR was already big enough without more refactoring :)

Prox/Prox/Data/PlacesController.swift
thebnich: Probably. As a short-term fix, I can make a public method that locks and calls this one.

The locking situation in `PlacesProvider` is a big mess; we should rewrite these critical sections using GCD instead.

Prox/Prox/Data/PlacesController.swift
```diff
@@ -139,6 +129,28 @@ class PlacesProvider {
...
+        let toRoots = CategoriesUtil.categoryToRootsMap
+
+        return allPlaces.filter { place in
+            let categories = Set(place.categories.ids.flatMap { toRoots[$0] }.reduce([], +))
+            return !enabledCategories.isDisjoint(with: categories)
+        }
+    }
+
+    /// Applies the current set of filters to all places, setting `displayedPlaces` to the result.
+    /// Callers must acquire a write lock before calling this method!
```
thebnich: > Are we locking whenever we access allPlaces? Add some comments (e.g. at the lock declaration) please!

I think so, but I didn't write this code, so I don't know exactly what `placesLock` is supposed to protect.

I created issue #529, which should clean up this code in general.

Prox/Prox/Data/PlacesController.swift
thebnich: Yeah, I see we're doing this elsewhere in the file:
```
self.placesLock.withReadLock {
    self.delegate?.placesProvider(self, didUpdatePlaces: self.displayedPlaces)
}
```

That means we're holding the read lock for the entire execution of the `didUpdatePlaces` callback, so We're Gonna Have A Bad Time. Specifically [this](https://github.com/mozilla-mobile/prox/blob/eb86ad2227be304457fb526a041ab8ddb04af613/Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift#L449).

Prox/Prox/PlaceDetails/FilterViewController.swift
thebnich: What do you mean by "dismiss a `UIAlert`"? How are `UIAlert`s related to this VC?

Prox/Prox/PlaceDetails/FilterViewController.swift
thebnich: Don't see perf issues, but I agree we may as well cache it.

Prox/Prox/Widgets/InsetButton.swift
```diff
@@ -0,0 +1,37 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override var intrinsicContentSize: CGSize {
+        let size = super.intrinsicContentSize
+        return CGSize(width: size.width + titleEdgeInsets.left + titleEdgeInsets.right,
+                      height: size.height + titleEdgeInsets.top + titleEdgeInsets.bottom)
+    }
+
+    @objc private func didTouchDismiss() {
```
thebnich: I'll just remove this from `InsetButton` now. In general, it doesn't look like we're really maintaining pressed states in Prox, so we can add this later with specific UX input.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
mcomella: I guess I didn't think through it. :P I'd imagine it'd be ^ or id-lookup, but that indirection adds some complexity so maybe it's not worth it.

Prox/Prox/Extensions/UIColorExtensions.swift
```diff
@@ -0,0 +1,31 @@
...
+private struct Color {
+    var red: CGFloat
+    var green: CGFloat
+    var blue: CGFloat
+};
+
+extension UIColor {
+    /**
+     * Initializes and returns a color object for the given RGB hex integer.
+     */
```
thebnich: Yeah, this was copy/pasted (and modified) from Firefox iOS. Apologies for the sloppy import!

Prox/Prox/PlaceDetails/FilterViewController.swift
mcomella: I'm concerned that `dismiss` can be used to dismiss a ViewController other than self – like a UIAlertViewController.

PR #527, opened 2017-02-16
No issue: rm most of carousel

mcomella:
I attempted to remove the place carousel but hit the point of diminishing returns. The VC still exists and is largely there to contain the PlacesProvider and forward the calls to the Details VC.

Also, I cleaned up some other random stuff.

This PR could be cleaner but I don't want to spend the time and I think it's incrementally better anyway.


Prox/Prox/Widgets/Prompts.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    // HACK: this variable. Maybe it can be replaced if we subclass UIAlertController and we compare
+    // the presented controller type.
+    static var isNoLocationAlertPresented = false
+    static func presentNoLocationAlert(for controller: UIViewController) {
+        // The message says to close and restart Prox, however, at time of writing, this is not
+        // always necessary: if we receive a location event while or after the dialog is displayed,
+        // we'll show the places. It's used as a catch all because we don't know for sure what
+        // causes the loading screen stall (#392).
+        let alert = UIAlertController(title: "Where did you go?",
```
thebnich: Just a note that I think we should start collecting these strings in a `Strings` struct (I created such a struct in the filtering PR). If we get to the point where we want to localize Prox, it'll be a chore to grep for every string and weed out the user-facing ones. Throwing these into a strings struct is a low-effort way to keep them separate (and we don't even need to worry about `NSLocalizedString` wrappers for now).

Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -38,7 +38,7 @@ fileprivate let scaleOutTransformRight = CGAffineTransform.identity.translatedBy
  **/
 class PlaceDetailViewController: UIViewController {
 
-    weak var dataSource: PlaceDataSource? {
+    weak var dataSource: PlacesProvider? {
```
thebnich: Since we already have the `PlaceProvider` interface used for `Place`s, do you think it would make sense to keep the `PlaceDataSource` name here instead of `PlacesProvider`? Alternatively, we could find a new name for `PlaceProvider`.

mcomella: `PlaceDataSource` wfm – I kept it as `PlacesProvider` for simplicity in not touching every file. I filed #528.

Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -234,15 +200,15 @@ extension PlaceCarouselViewController: InAppNotificationToastDelegate {
...
+        if !Prompts.isNoLocationAlertPresented {
             updateLocation(location: location)
             return
         }
 
         // If we don't dismiss the location alert, the detail view controller (called from
         // updateLocation) does not get presented.
         dismiss(animated: true) {
-            self.isNoLocationAlertPresented = false
+            Prompts.isNoLocationAlertPresented = false
```
mcomella: It manages it if the close button is pressed but not if it's dismissed in code (as it is here). We could probably fix this by subclassing UIAlertViewController & overriding `viewWillDisappear` or whatever.

Prox/Prox/Widgets/Prompts.swift
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import UIKit
+
+/// A collection of random prompt functions.
+/// Maybe these would be better as a UIAlertController subclass. If so, make sure you continue to
+/// log the analytics in a way people won't forget to call it (e.g. in `viewWillAppear`).
+struct Prompts {
```
mcomella: > Is your concern that putting it in the file directly makes that file too big?

Precisely. I've been finding it's much easier to quickly & confidently edit a file if all of the things that don't concern what I'm trying to do are encapsulated away into their own files (e.g. these Prompts have nothing to do with an particular VC) & the file isn't too large.

> A generic prompt container doesn't really make sense as under Widget/ to me, where I consider "widget" to be a general-purpose, reusable UIView subclass. Are you using widgets for different things in Prox?

That's generally what I consider a widget but I didn't know where else to put these. Perhaps a top-level Prompts file or `/Prompts/` with a separate file for each?

PR #545, opened 2017-02-24
Follow-up: Mock tweak update

liuche:
After trying it out on an iPhone 7 Plus, updated mock to include additional spacing of 10.


liuche:
In-person review by @mcomella

Issue #544, opened 2017-02-24
Transition map view controller in from top


mcomella:

mcomella:
Note to self: might be able to copy this animation from the FilterViewController.

PR #546, opened 2017-02-24
No issue: map view updates with places

mcomella:
Still TODO:
* ~~Hide card until place is selected~~ (followup #548)
* ~~How to handle map when place is selected? (return false from delegate?)~~
* ? When clicking card, go to detail CV or new view in mocks (#550)
* Transition down from top (#544)
* Use our custom dot (#547)
* Do circle inside circle view (#551)
* Clustering? (#549)


Prox/Prox/Widgets/ExpandingCardView.swift
```diff
@@ -0,0 +1,135 @@
...
+    }
+    public var shadowOpacity: Float {
+        get { return layer.shadowOpacity }
+        set { layer.shadowOpacity = newValue }
+    }
+    public var shadowRadius: CGFloat {
+        get { return layer.shadowRadius }
+        set { layer.shadowRadius = newValue }
+    }
+    public var shadowOffset: CGSize {
```
thebnich: Any specific reason you're exposing these `layer` properties as properties on the view? If I were using this class, I would expect to do `expandingCardView.layer.shadowColor = ...` like any other `UIView`.

Prox/Prox/Widgets/ExpandingCardView.swift
```diff
@@ -0,0 +1,135 @@
...
+        constraint.priority = UILayoutPriorityRequired - 1
+        return constraint
+    }()
+
+    /// A sequence of constraints from the current user-defined contentView.
+    private var contentViewConstraints: [NSLayoutConstraint] = []
+
+    private lazy var scrollView: UIScrollView = {
+        let view = ObservableScrollView()
+        view.onContentSizeUpdated = { [unowned self] contentSize in
```
thebnich: Note that this is typically done use the delegate pattern (`ObservableScrollViewDelegate` with `didUpdateContentSize`, for example). That way, the client code isn't responsible for breaking the ref cycle, which is error prone.

Prox/Prox/Map View/MapViewCardFooter.swift
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import UIKit
+import SnapKit
+
+private let cardPadding = 10
+private let cardSpacing: CGFloat = 10
```
thebnich: Should these go into the `Style` struct you made?

Prox/Prox/Map View/MapViewCardFooter.swift
```diff
@@ -5,8 +5,13 @@
...
 
-private let cardPadding = 10
-private let cardSpacing: CGFloat = 10
+private let cardHorizontalPadding = 16
+private let cardBottomPadding: CGFloat = 24
+
+private let cardSpacing = 18
+private let cardProviderSpacing = 12
+
+private let cardCoverPhotoHeight = 72
```
thebnich: Same question here re: `Style`.

Prox/Prox/Map View/MapViewReviewProvider.swift
```diff
@@ -0,0 +1,65 @@
...
+        let view = UIImageView()
+        view.clipsToBounds = true
+        view.contentMode = .scaleAspectFit
+        view.image = #imageLiteral(resourceName: "score_yelp_5")
+        return view
+    }()
+
+    private lazy var reviewCountView: UILabel = {
+        let view = UILabel()
+        view.text = "10 reviews"
```
thebnich: Move this to `Strings`?

Prox/Prox/Map View/MapViewCardFooter.swift
```diff
@@ -43,8 +43,8 @@ class MapViewCardFooter: ExpandingCardView {
         return view
     }()
 
-    private lazy var yelpProviderView: UIView = MapViewReviewProvider()
-    private lazy var tripAdvisorProviderView: UIView = MapViewReviewProvider()
+    private lazy var yelpProviderView: UIView = MapViewYelpReviewProviderView()
```
thebnich: :trollface:

mcomella: :|

Prox/Prox/Map View/MapViewCardFooter.swift
mcomella: Global constants/strings/etc. create a lot of bloat in files until it's hard to maintain, easy to loose track of unused values, and difficult to identify what is safe to re-use (and what just happens to be in a giant constants file). I prefer to keep the constants locally to the places they're used. A good example is our [FFAndroid styles](https://dxr.mozilla.org/mozilla-central/source/mobile/android/base/resources/values/styles.xml) which are so hard to maintain (most of these styles should really be embedded directly in the layout files). fwiw, I'm conflicted about our Colors.swift & Fonts.swift files.

Style.swift was intended to be for shared styles – should I rename it as such?

Prox/Prox/Widgets/ExpandingCardView.swift
mcomella: No – fixed.

Prox/Prox/Map View/MapViewReviewProvider.swift
mcomella: Re-used our "no info" text.

Prox/Prox/Widgets/ExpandingCardView.swift
mcomella: Fixed.

PR #555, opened 2017-02-24
Closes #511: refine groups for filtering

### LINKED ISSUE #511: Refine groups for filtering
Not sure if we filed this anywhere but here's the category grouping we want to start using based on what we get back from Yelp right now.

**Discover**
- Active Life
- Arts & Entertainment
- Local Flavor
- Hotels & Travel
- Landmarks & Historical Buildings **(Sub-category under Public Services & Government)**

**Eat & Drink**
- Food
- Nightlife
- Restaurants

**Shop**
- Shopping

**Services**
- Automotive
- Beauty & Spas
- Bicycles
- Education
- Event Planning & Services 
- Financial Services
- Health & Medical
- Home Services
- Local Services
- Professional Services
- Mass Media
- Pets
- Public Services & Government 
- Real Estate
- Religious Organizations

thebnich:
Examining the list of places in each filter, we've noticed that the current partitioning isn't very well defined. In particular, limo rentals and gyms are showing up in Discover, while unique landmarks are being buried under Services.

To fix this properly, we need to allow more fine-grained category -> filter mapping since we currently only categorize the roots.

### DIFF
--- a/Prox/Prox/PlaceDetails/FilterViewController.swift
+++ b/Prox/Prox/PlaceDetails/FilterViewController.swift
@@ -5,8 +5,8 @@
 import Foundation
 
 protocol FilterViewControllerDelegate: class {
-    func filterViewController(_ filterViewController: FilterViewController, didUpdateFilters enabledFilters: Set<Filter>, topRatedOnly: Bool)
-    func filterViewController(_ filterViewController: FilterViewController, didDismissWithFilters enabledFilters: Set<Filter>, topRatedOnly: Bool)
+    func filterViewController(_ filterViewController: FilterViewController, didUpdateFilters enabledFilters: Set<PlaceFilter>, topRatedOnly: Bool)
+    func filterViewController(_ filterViewController: FilterViewController, didDismissWithFilters enabledFilters: Set<PlaceFilter>, topRatedOnly: Bool)
 }
 
 /// A drop-down view from the top of the screen that displays a list of categories to filter.
@@ -18,18 +18,18 @@ class FilterViewController: UIViewController {
 
     fileprivate let background = UIButton()
     private let stackView = UIStackView()
-    private(set) var enabledFilters: Set<Filter>
+    private(set) var enabledFilters: Set<PlaceFilter>
     private let placeCountLabel = UILabel()
     private let topRatedSwitch = UISwitch()
 
-    private let filterLabels: [Filter: String] = [
-        .discover: Strings.filterView.discover,
-        .eatAndDrink: Strings.filterView.eatAndDrink,
-        .shop: Strings.filterView.shop,
-        .services: Strings.filterView.services,
+    private let filterLabels: [(PlaceFilter, String)] = [
+        (.discover, Strings.filterView.discover),
+        (.eatAndDrink, Strings.filterView.eatAndDrink),
+        (.shop, Strings.filterView.shop),
+        (.services, Strings.filterView.services),
     ]
 
-    init(enabledFilters: Set<Filter>, topRatedOnly: Bool) {
+    init(enabledFilters: Set<PlaceFilter>, topRatedOnly: Bool) {
         self.enabledFilters = Set(enabledFilters)
         super.init(nibName: nil, bundle: nil)
 
@@ -147,7 +147,7 @@ class FilterViewController: UIViewController {
     }
 
     @objc private func didToggleFilter(sender: FilterButton) {
-        let filter = Filter(rawValue: sender.tag)!
+        let filter = PlaceFilter(rawValue: sender.tag)!
         enabledFilters.formSymmetricDifference([filter])
         sender.isSelected = enabledFilters.contains(filter)
         delegate?.filterViewController(self, didUpdateFilters: enabledFilters, topRatedOnly: topRatedSwitch.isOn)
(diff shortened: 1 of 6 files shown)


Prox/Prox/Utilities/CategoriesUtil.swift
```diff
@@ -124,6 +124,33 @@ struct CategoriesUtil {
...
+
+        return categoryToFilter
+    }()
+
+    /// Returns this category, its root categories, and all parents in between.
+    /// The result is DFS-ordered.
+    private static func getAncestors(forCategory category: String) -> [String] {
+        let ancestors = categoryToParentsMap[category] ?? []
+        let hierarchy = ancestors.map { getAncestors(forCategory: $0) }.reduce([category], +)
+        return Array(NSOrderedSet(array: hierarchy)) as! [String]
```
mcomella: Why can't you return `hierarchy` directly?

Prox/Prox/Utilities/CategoriesUtil.swift
```diff
@@ -124,6 +124,33 @@ struct CategoriesUtil {
...
+        for category in categoryToParentsMap.keys {
+            guard let firstCategorizedParent = getAncestors(forCategory: category).flatMap({ categoryToFilter[$0] }).first else { continue }
+            categoryToFilter[category] = firstCategorizedParent
+        }
+
+        return categoryToFilter
+    }()
+
+    /// Returns this category, its root categories, and all parents in between.
+    /// The result is DFS-ordered.
```
mcomella: nit: Weird to phrase it as DFS because the tree is conceptually upside down and, afaik, we don't intentionally give the child ordered parents.

Prox/Prox/Utilities/CategoriesUtil.swift
```diff
@@ -124,6 +124,33 @@ struct CategoriesUtil {
...
+        // Build the initial map of categories -> filters that we set explicitly.
+        for (filter, categories) in Filter.categories {
+            for category in categories {
+                categoryToFilter[category] = filter
+            }
+        }
+
+        // Set the remaining categories based on the first ancestor with an explicitly-set filter.
+        for category in categoryToParentsMap.keys {
+            guard let firstCategorizedParent = getAncestors(forCategory: category).flatMap({ categoryToFilter[$0] }).first else { continue }
```
mcomella: Is it ever expected that a category will not have a categorized ancestor? If so, explain why not (on the outer `categoryToFilter` var?). If not, we should crash (since this should be caught by tests. Also, write a test :) ).

Prox/Prox/Utilities/CategoriesUtil.swift
mcomella: The results is DFS-ordered but our Yelp categories are not guaranteed to be returned in any order, right? afaik, Yelp doesn't put the parents array in any particular order and could resort them when the file is updated. This means our behavior could change if we add a new Yelp categories list.

I think we should add a sort in our `categoryToParentsMap` init code.

Prox/Prox/PlaceDetails/FilterViewController.swift
```diff
@@ -18,14 +18,19 @@ class FilterViewController: UIViewController {
...
-    private let filters: [PlaceFilter]
-    private(set) var enabledFilters: [Bool]
+    private(set) var enabledFilters: Set<Filter>
     private let placeCountLabel = UILabel()
     private let topRatedSwitch = UISwitch()
 
-    init(filters: [PlaceFilter], enabledFilters: [Bool], topRatedOnly: Bool) {
-        self.filters = filters
-        self.enabledFilters = enabledFilters
+    private let filterLabels: [Filter: String] = [
```
thebnich: Thanks for mentioning ordering -- I actually had this as an array of (enum, label) tuples before this PR, then "cleaned" it up to be a dictionary. But your comment reminds me of why I didn't use a dictionary in the first place: we lose the ordering! So I'll probably go back to tuples here.

My reasoning for keeping this in FVC is that strings are user-facing, so I prefer the presentation layer to own them rather than mixing them with the model.

Prox/Prox/Utilities/CategoriesUtil.swift
thebnich: > Is it ever expected that a category will not have a categorized ancestor?

Yes: the categories come from Firebase, which comes from crawling Yelp, which we know can update the categories (which is how I found we had a mismatch in the first place).

Prox/Prox/Utilities/CategoriesUtil.swift
thebnich: How about "DFS-ordered from this category to its roots"?

thebnich: Anthony and I were looking through the categories yesterday and found that the first category often seems to best represent a place. Perhaps that was luck? I hope not since we now associate places with a single filter, and we wouldn't have a good way to choose which parent to use.

Prox/Prox/Utilities/CategoriesUtil.swift
thebnich: `NSOrderedSet` removes duplicates (which isn't strictly required with how we're using the result, but seemed like a good idea anyway).

Prox/Prox/Utilities/CategoriesUtil.swift
thebnich: Since this has a significant impact on which filters will show which places, and since the best parent varies by category, I think we'll want to create an "override" for each category that has multiple parents. It looks like there are only 20 or so such categories, so we can figure these out manually.

This requires some product involvement, so I'm going to punt on this for now and address in #559.

PR #561, opened 2017-02-27
No issue: Don't display missing location dialog when location faking.

mcomella:


Prox/Prox/PlaceCarousel/PlaceCarouselViewController.swift
```diff
@@ -144,7 +144,12 @@ extension PlaceCarouselViewController: LocationMonitorDelegate {
 
     func locationMonitor(_ locationMonitor: LocationMonitor, didFailInitialUpdateWithError error: Error) {
         // Note: actually, at this point, if the location comes, `didUpdateLocation` will be called.
-        Prompts.presentNoLocationAlert(for: self)
+        if AppConstants.MOZ_LOCATION_FAKING,
+                let location = locationMonitor.getCurrentLocation() { // should never be nil.
+            self.locationMonitor(locationMonitor, didUpdateLocation: location)
+        } else {
+            Prompts.presentNoLocationAlert(for: self)
+        }
```
thebnich: Since this isn't really an error, I think would make more sense to "redirect" this callback internally in `LocationMonitor`'s `didFailWithError`. That way clients implementing this delegate (i.e., `PlaceCarouselViewController`) don't have to handle this condition themselves.

PR #572, opened 2017-02-28
No issue: rip out events code

thebnich:
Given the following:
1) the events code is not currently being used, therefore maintained,
2) none of the current Prox engineers have a strong understanding of the current events code, and
3) (IIRC) the events code wasn't terribly reliable,

I suggest we remove events/notifications altogether.


Prox/Prox/AppDelegate.swift
```diff
@@ -61,7 +49,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func applicationDidEnterBackground(_ application: UIApplication) {
         // if there is a timer running, cancel it. We'll wait until background app refresh fires instead
         locationMonitor.cancelTimeAtLocationTimer()
```
mcomella: I think this "time at location" was so that we would fetch events if the user stayed at a place long enough. I think we can remove it.

Prox/Prox/RemoteConfigKeys.swift
```diff
@@ -49,30 +49,6 @@ class RemoteConfigKeys {
     // This is measure in minutes.
     public static let maxWalkingTimeInMins = RemoteConfigInt(key: "max_walking_time_in_mins", defaultValue: 30)
 
-    // The event search radius the app will use to query the Firebase datbase
-    // This is measures in kilometers.
-    public static let eventSearchRadiusInKm = RemoteConfigRadius(key: "event_search_radius_in_km", defaultValue: 40.0)
```
mcomella: Can you rm the values from Firebase too?

PR #574, opened 2017-02-28
Closes #526: filter view card animation

### LINKED ISSUE #526: Improve filtering animation
Right now, the filter UI just overlays the top of the card. The mockups show an animation that pushes the card down, so we should implement that.

thebnich:


Prox/Prox/PlaceDetails/PlaceDetailViewController.swift
```diff
@@ -683,6 +697,12 @@ class PlaceDetailViewController: UIViewController {
         filterVC.delegate = self
         filterVC.placeCount = dataSource.numberOfPlaces()
         present(filterVC, animated: true, completion: nil)
+
+        UIView.animate(withDuration: cardSlideDuration) {
```
mcomella: nit: could be a helper method, `slideCurrentCardView(willBeShown:)`. That way, there's no code duplication (between hide & show) and it's a little more readable.

PR #575, opened 2017-03-01
No issue: Fix Xcode schemes

thebnich:
This removes the `Release` and `Debug` schemes, leaving us with just `Prox Mock Location` and `Prox Current Location`. I added `Mock Location (Debug)` and `Current Location (Debug)` build configurations, so the mock/current location schemes now correctly switch between the corresponding configuration based on the action (Run/Test/Archive/etc). In other words, "debug" builds are no longer mutually exclusive from each build channel, and each build channel has a debug variant.

As @sleroux suggested, this switches all schemes to use the Enterprise team. That means you should be able to run directly on your iPhone from Xcode if you're in the Mozilla Enterprise group.


Prox/Prox/Utilities/AppConstants.swift
```diff
@@ -32,18 +30,21 @@ public struct AppConstants {
     public static let BuildChannel: AppBuildChannel = {
         #if MOZ_CHANNEL_CURRENT_LOCATION
             return AppBuildChannel.CurrentLocation
-        #elseif MOZ_CHANNEL_RELEASE
-            return AppBuildChannel.Release
-        #elseif MOZ_CHANNEL_MOCK_LOCATION
-            return AppBuildChannel.MockLocation
         #else
-            return AppBuildChannel.Debug
+            #if !MOZ_CHANNEL_MOCK_LOCATION
```
mcomella: nit: don't waste time changing now but I think it'd be clearer as

if CURRENT_LOC
else if MOCK_LOC
else
  assertionfailure

Prox/Prox/Utilities/Logger.swift
```diff
@@ -26,16 +26,7 @@ class Logger {
...
-        switch AppConstants.BuildChannel {
-        case .Debug:
-            minPrintedLevel = .debug
-        case .Release:
-            minPrintedLevel = .warn
-        case .CurrentLocation: fallthrough
-        case .MockLocation:
-            minPrintedLevel = .info
-        }
+        let minPrintedLevel: LoggerLevel = AppConstants.isDebug ? .debug : .warn
```
mcomella: Any reason we shouldn't log everything always?

PR #576, opened 2017-03-02
Closes #560: Reads website urls from Google provider.

### LINKED ISSUE #560: Places should have website data

liuche:
Fetches website from new Google provider on server.


Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -242,7 +242,7 @@ class PlaceDetailsCardView: UIView {
         // Labels will gracefully collapse on nil.
         titleLabel.text = place.name
         categoryLabel.text = PlaceUtilities.getString(forCategories: place.categories.names)
-        updateURLText(place.url)
+        updateURLText(place.googleProvider?.url)
```
thebnich: This PR shouldn't have to touch `PlaceDetailsCardView` or `PlaceDetailsCardViewController` IMO -- if we want to prioritize Google URLs over others, we can set that URL to the place URL.

Prox/Prox/Models/PlaceProviders.swift
```diff
@@ -62,7 +62,7 @@ class SinglePlaceProvider: PlaceProvider {
         let photoURLStrings = (dict["images"] as? [[String:String]])?.flatMap { $0["src"] } ?? []
         photoURLs = photoURLStrings.flatMap { URL(string: $0) }
 
-        if let urlString = dict["url"] as? String,
+        if let urlString = dict["url"] as? String ?? dict["website"] as? String,
```
thebnich: IIRC, `website` is going to the be the "real" website (if it exists), whereas `url` will be the provider-specific website. If so, invert the order here?

It might also be worth having a separate `website` field in `PlaceProvider`, but I suppose we can add that when we need it.

Prox/Prox/Models/Place.swift
```diff
@@ -90,7 +94,12 @@ class Place: Hashable {
             wikipediaProvider = SinglePlaceProvider(fromDictionary: wikipediaDict)
         }
 
-        let providers: [PlaceProvider?] = [yelpProvider, tripAdvisorProvider, wikipediaProvider]
+        var googleProvider: SinglePlaceProvider?
+        if let googleDict = details.childSnapshot(forPath: GOOGLE_PATH).value as? [String: Any] {
+            googleProvider = SinglePlaceProvider(fromDictionary: googleDict)
+        }
+
+        let providers: [PlaceProvider?] = [yelpProvider, tripAdvisorProvider, wikipediaProvider, googleProvider]
```
thebnich: I think `googleprovider` should be first (see other comment about priority).

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
liuche: This is a case where we either want to use the google url, or none at all - I don't think we want to show the Yelp url at all, because we don't want to emphasize Yelp more, and there are already 2 other entry points to the Yelp url.

liuche:
As a follow-up to review comments, we're differentiating between "url" (which is a provider url) and "website" which is the website of the location, e.g., not inside of Yelp. This adds the website field to the providers, and I've tested it with the updated prox-server "website" schema.

PR #577, opened 2017-03-02
Closes #569: Integrate hand-curated Discover data with client

### LINKED ISSUE #569: Integrate hand-curated Discover data with client

thebnich:
Two small commits that:
1) fix Discover to only list our hand-curated `proxdiscover` IDs, and
2) include the `custom` provider to generate Places.

### DIFF
--- a/Prox/Prox/Utilities/CategoriesUtil.swift
+++ b/Prox/Prox/Utilities/CategoriesUtil.swift
@@ -131,8 +131,15 @@ struct CategoriesUtil {
         var categoryToFilter = [String: PlaceFilter]()
 
         // Build the initial map of categories -> filters that we set explicitly.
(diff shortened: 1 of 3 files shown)


Prox/Prox/Utilities/CategoriesUtil.swift
```diff
@@ -131,8 +131,12 @@ struct CategoriesUtil {
         var categoryToFilter = [String: PlaceFilter]()
 
         // Build the initial map of categories -> filters that we set explicitly.
-        for (filter, categories) in PlaceFilter.categories {
+        for (var filter, categories) in PlaceFilter.categories {
             for category in categories {
+                // Remap Discover -> Services for now.
```
mcomella: nit: Might want to explain a bit more, in case this code lasts longer than we'll remember what this meant.

PR #579, opened 2017-03-02
Closes #578: don't show places that won't be open today or tomorrow

### LINKED ISSUE #578: Don't show places that won't be open today or tomorrow
If we have times for a place, and the times indicate that the place isn't open now and won't be open later today, we should omit the place from the list.

thebnich:


thebnich:
Note: going to wait until next week until more events have been added and we can determine what we want this to look like.

thebnich:
Updated to hide events that aren't open today or tomorrow.

Prox/Prox/Models/Place.swift
```diff
@@ -423,7 +423,7 @@ struct OpenHours {
 
     func nextOpeningTime(forTime baseTime: Date) -> String? {
         let midnight = DateComponents(hour: 0, minute: 0)
-        let times = [baseTime] + (1...7).map { getDate(forTime: midnight, onDate: baseTime.addingTimeInterval(TimeInterval($0) * AppConstants.ONE_DAY))! }
+        let times = [baseTime] + (1...1).map { getDate(forTime: midnight, onDate: baseTime.addingTimeInterval(TimeInterval($0) * AppConstants.ONE_DAY))! }
```
mcomella: nit: Either:
1) Update the method name to make it clear the span for which it turns `nextOpeningTime` (today & tomorrow)
2) Add a comment that we hacked this to only return today and tomorrow

PR #596, opened 2017-03-03
Closes #583: hide rating views if no Yelp or TA

### LINKED ISSUE #583: Hide ratings on details view when not available

thebnich:


Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -63,9 +63,7 @@ struct PlaceUtilities {
     static func updateReviewUI(fromProvider provider: PlaceProvider?, onView view: ReviewContainerView, isTextShortened: Bool = false) {
         guard let provider = provider,
                 !(provider.totalReviewCount == 0 && provider.rating == nil) else { // intentional: if both null, short-circuit
-            setSubviewAlpha(0.4, forParent: view)
```
mcomella: What if 1 provider has data but the other does not? Do we really want to one provider's collapse the UI in that case?

Prox/Prox/Widgets/ReviewContainerView.swift
```diff
@@ -106,6 +115,8 @@ class ReviewContainerView: UIView {
 
         constraints += [bottomAnchor.constraint(equalTo: numberOfReviewersLabel.bottomAnchor)]
 
+        collapsedConstraints = [ heightAnchor.constraint(equalToConstant: 0) ]
```
mcomella: If we set a height constraint of 0, then disable it, is the view going to want to stay size 0? Or does it go back and use the intrinsic content size again? I'm assuming the latter, but just want to make sure you check this case!

Prox/Prox/Utilities/PlaceUtilities.swift
thebnich: Oops. I saw the "if both null" comment and assumed "both" meant TA and Yelp.

Prox/Prox/Widgets/ReviewContainerView.swift
thebnich: I'm not sure TBH, but does it matter? Don't we create a new view for every place?

thebnich: I'll add a comment mentioning this as a shortcoming of the `isCollapsed` setter if you're just concerned about future code.

thebnich:
Updated. Verified that switching to/from hidden reviews works fine when swiping between cards.

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -271,6 +274,10 @@ class PlaceDetailsCardView: UIView {
 
         PlaceUtilities.updateReviewUI(fromProvider: place.yelpProvider, onView: yelpReviewView)
         PlaceUtilities.updateReviewUI(fromProvider: place.tripAdvisorProvider, onView: tripAdvisorReviewView)
+
+        let collapsed = (place.totalReviewCount == 0)
+        reviewViewContainer.isHidden = collapsed
```
mcomella: Why do we need both height == 0 and isHidden?

thebnich: `height == 0` will set the container constraints so that the top and bottom are the same, but subviews (images) might still be visible. So `isHidden` makes the view disappear, and `height == 0` collapses the layout.

PR #604, opened 2017-03-05
Closes #587: Use categories of "proxdiscover-" items directly

### LINKED ISSUE #587: Discover is missing category descriptions
It's probably because the alias -> text lookup failed since we're not using official yelp categories.

liuche:
This special-cases the categories for proxdiscover- items by checking for the proxdiscover- id prefix if category matching fails.


Prox/Prox/Models/PlaceProviders.swift
```diff
@@ -47,9 +47,14 @@ class SinglePlaceProvider: PlaceProvider {
         let categoryIds = dict["categories"] as? [String] ?? []
         var categories = (names: [String], ids: [String])([], [])
         for id in categoryIds {
-            guard let name = CategoriesUtil.categoryToName[id] else { continue }
-            categories.ids.append(id)
-            categories.names.append(name)
+            if let name = CategoriesUtil.categoryToName[id] {
+                categories.ids.append(id)
+                categories.names.append(name)
+            } else if let placeid = self.id, placeid.hasPrefix("proxdiscover-") {
```
mcomella: nit: we should probably make "proxdiscover-" a constant considering it's used twice.

PR #610, opened 2017-03-07
Closes #563: future planning in map view

### LINKED ISSUE #563: Implement search when panning in Map View
prox-search-move-i04

When the user pans around the Map View:

* A button labelled “Search Here” should appear underneath the map

When tapped, a search is performed around the selected radius:

* The button changes label to “Searching…” and greys out 
* User can no longer pan or zoom around the map
* If any card is currently selected, it should animate out and be deselected

When search has been performed:

* The button disappears
* Map pins/place indicators animate in (bounce) from the top
* **Seen on animation, but don’t do it**: the closest place on the map is selected

mcomella:
#563 

I didn't have enough time to fully finish but I think this is good enough:

1. There is no "reset" button in dialog
1. When the user clicks "Search here", it only finds places in our initial search radius, not a radius calculated for the current zoom level
1. The code is pretty hairy

@antlam I'm most concerned about 2. What do you think?


mcomella:
Crash landed on master so it's ready for user testing – @thebnich @liuche please still review and we can make any essential changes tomorrow.

PR #614, opened 2017-03-08
Closes #608 - Lists event times as day + time

### LINKED ISSUE #608: Event cards should have better opening time information
While we might not have time to build a distinct design for event cards, we can still alter them slightly so they’re more helpful.

Instead of saying “Closed until 18:00”, we can say “18:00 Start time”. Events aren’t closed or open like venues are, so putting a start time is more accurate, and all we need to do is to alter our existing copy.

modal

liuche:
Adds an event times function to the OpenHours class - it's a little hacky because it highly resembles getNextOpeningHours, but refactoring more didn't seem like it had clear benefits, and would touch a lot of code.


Prox/Prox/Models/Place.swift
```diff
@@ -451,6 +451,37 @@ struct OpenHours {
         return nil
     }
 
+    func getEventTimeText(forTime baseTime: Date) -> (String, String)? {
+        // Note: this function is basically a modified version of nextOpeningTime()
+        let midnight = DateComponents(hour: 0, minute: 0)
+
+        // HACK: Keeping this as a range so we can easily change this function look ahead more days if needed.
+        // (1...N) will look ahead N days after today.
+        let times = [baseTime] + (1...1).map { getDate(forTime: midnight, onDate: baseTime.addingTimeInterval(TimeInterval($0) * AppConstants.ONE_DAY))! }
```
mcomella: nit: -> `todayAndTomorrow`, or something. It's unclear which times these are and how we intend to use them.

Prox/Prox/Models/Place.swift
```diff
@@ -451,6 +451,37 @@ struct OpenHours {
         return nil
     }
 
+    func getEventTimeText(forTime baseTime: Date) -> (String, String)? {
```
mcomella: nit: -> `getEventTimeText(forToday todayDate: Date)`

This method returns "Today" if the given date has an opening date, making the output dependent on the current date - we should make that clear.

Prox/Prox/Models/Place.swift
```diff
@@ -451,6 +451,37 @@ struct OpenHours {
...
+        let times = [baseTime] + (1...1).map { getDate(forTime: midnight, onDate: baseTime.addingTimeInterval(TimeInterval($0) * AppConstants.ONE_DAY))! }
+
+        for time in times {
+            guard let allOpeningTimes = getOpeningTimes(forDate: time),
+                  let openingPeriod = getCurrentOpeningPeriod(fromOpeningPeriods: allOpeningTimes, forDate: time) ?? getNextOpeningPeriod(fromOpeningPeriods: allOpeningTimes, forDate: time) else {
+                continue
+            }
+
+            let openingDayText: String
+            if OpenHours.calendar.isDate(time, inSameDayAs: baseTime) {
```
mcomella: This is always true for index 0, the next if is always true for index 1, and else will be called for the rest – this makes the code smell to me.

That being said, in this implementation, there is a separation from the array ordering and the return value (but this abstraction adds complexity and confuses the reader about motivations! :) ).

I think there is a clearer solution here but I can't think of it off the top of my head so this is probably good enough. Maybe add a comment to say it can be clearer?

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -319,7 +319,7 @@ class PlaceDetailsCardView: UIView {
             return
         }
 
-        guard let (primaryText, secondaryText) = getStringsForOpenHours(hours, forDate: Date()) else {
+        guard let (primaryText, secondaryText) = getStringsForOpenHours(hours, forDate: Date(), isEvent: place.customProvider?.description != nil) else {
```
mcomella: It's slightly safer to additionally check: `!descrption.isEmpty`, just in case we screw up the schema.

liuche:
Yeah, I didn't feel like I wanted to refactor all that, and when I initially tried that approach, it involved de-private-ing a lot of functions that still didn't do exactly what we wanted. Tried to minimize code change here, but filing #618 to clean this up later. (Maybe next week, when we have LOTS OF TIME? :D)

Prox/Prox/Models/Place.swift
thebnich: We could do `in times.enumerate()`, then switch on the index rather than doing the `inSameDayAs` checks.

liuche: I think this should be part of a refactor, returning a date and/or time object, and then allowing the consumer (PlaceDetails) to format the result however it needs. This would certainly fix the problem both here and in nextOpeningTime.

PR #616, opened 2017-03-08
Closes #543: add timeout when calculating travel ETAs

### LINKED ISSUE #543: Device hangs when retrieving many places
I think this may be related to waiting for `Promise.allFilled` for travel times.

thebnich:
I found that `MKDirections.calculateETA` simply isn't executing its callback at all, which seems like an iOS bug. I added a delayed dispatch on the main thread to see if maybe the main thread is deadlocked and we're seeing this as a side effect, but the dispatched callback executes just fine after the app freezes.

Was originally going to sort by raw distance and update the walking distance only as the cards are shown, but that would slightly reorder the cards, which is risky given the closeness to Chicago. Adding a timeout is probably the safest bet for now since it doesn't change the experience.

### DIFF
--- a/Prox/Prox/Utilities/PlaceUtilities.swift
+++ b/Prox/Prox/Utilities/PlaceUtilities.swift
@@ -28,7 +28,7 @@ struct PlaceUtilities {
     static func sort(places: [Place], byTravelTimeFromLocation location: CLLocation, ascending: Bool = true, completion: @escaping ([Place]) -> ()) {
         var sortedPlaces = PlaceUtilities.sort(places: places, byDistanceFromLocation: location)
         // this means we will probably only get walking directions for places, but I think that will be OK for now
-        PlaceUtilities.getTravelTimes(forPlaces: sortedPlaces, fromLocation: location, withTransitTypes: [.walking]).upon { result in
+        PlaceUtilities.updateTravelTimes(forPlaces: sortedPlaces, fromLocation: location, withTransitTypes: [.walking]).upon {
             let sortedByTravelTime = places.sorted { (placeA, placeB) -> Bool in
                 let placeAETA = PlaceUtilities.lastTravelTimes(forPlace: placeA)?.getShortestTravelTime() ?? Double.greatestFiniteMagnitude
                 let placeBETA = PlaceUtilities.lastTravelTimes(forPlace: placeB)?.getShortestTravelTime() ?? Double.greatestFiniteMagnitude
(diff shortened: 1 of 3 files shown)


Prox/Prox/Utilities/Future.swift
```diff
@@ -0,0 +1,18 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Deferred
+
+extension FutureProtocol {
+    /// Returns a Future resolved with the result of this Future, or nil
+    /// if this Future isn't resolved by the deadline.
+    func timeout(deadline: DispatchTime) -> Future<Value?> {
```
thebnich: Hm, I haven't run into any problems, though I haven't run this build much. Do you remember what kinds of problems you were seeing? I'd expect maybe compiler errors at worst, but this compiles fine.

Prox/Prox/Utilities/PlaceUtilities.swift
```diff
@@ -56,8 +56,10 @@ struct PlaceUtilities {
         }
     }
 
-    private static func getTravelTimes(forPlaces places: [Place], fromLocation location: CLLocation, withTransitTypes transitTypes: [MKDirectionsTransportType]) -> Future<[DatabaseResult<TravelTimes>]> {
-        return places.map { $0.travelTimes(fromLocation: location, withTransitTypes: transitTypes) }.allFilled()
+    private static func updateTravelTimes(forPlaces places: [Place], fromLocation location: CLLocation, withTransitTypes transitTypes: [MKDirectionsTransportType]) -> Future<Void> {
+        // HACK: travelTimes(fromLocation: location) may not fill its Deferred due to an apparent bug where
+        // MKDirections.calculateETA() never executes its callback. We set a timeout as a workaround.
+        return places.map { $0.travelTimes(fromLocation: location, withTransitTypes: transitTypes) }.allFilled().timeout(deadline: .now() + 5).ignored()
```
thebnich: The `places.map { ... }.allFilled().timeout(...)` will return a `Future<[DatabaseResult<TravelTimes>]?>`, so the `ignored()` converts this to `Future<Void>` to match the function return type.

I changed `getTravelTimes` -> `updateTravelTimes`, where we now return a `Future<Void>` instead of `Future<[DatabaseResult<TravelTimes>]>`. I did that because 1) we weren't using the result and 2) because the method also updates each `Place`, and side effects for getters are bad.

Prox/Prox/Utilities/Future.swift
mcomella: I think it was crashes. From #29: "When Place creation returns a nil, the app is killed deliberately causing all hell to break loose. Make this case handle more gracefully somehow"

And from the commit (https://github.com/mozilla-mobile/prox/commit/404217f754a45cb73a21da3a6ecdd29771fd1d6f#diff-ef52f2a05b75e3e96af7cc9c6dc8245eR7):
+// This is an empty class that forms a base class for all places
+// It is done this way because:
+// 1. Deferred cannot handle optionals
+// 2. Deferred cannot handle Protocols, even if they are class protocols
+// 3. Deferred can only handle actual classes. Therefore if a Venue fails to initalize
+//    because it doesn't have all the info it needs to make itself
+//    we cannot return a nil through the deferred, we have to return a result object
+//    that can encapsulate the actual result of the call while still returning a 
+//    guarenteed value for Deferred

Prox/Prox/Utilities/PlaceUtilities.swift
mcomella: > and side effects for getters are bad.

I like the rename. I regret that we're not doing this functionally though :P

thebnich:
> Only major concern <snip>

Then I think those places will be put at the end of the list, unsorted (which I think is what happens now when `MKDirections.calculateETA` starts throwing errors). I'm looking into a fix that _always_ tries to use raw distance as a fallback to make this better, but it's more complicated than expected, so I'll land that as a follow-up.

PR #624, opened 2017-03-08
Closes #620: Show toast explaining popularity sort

### LINKED ISSUE #620: Resolve behavior of "Sort by popularity" & Discover

thebnich:
Shows a toast the first time the user turns on popularity sort.

### DIFF
--- a/Prox/Prox/PlaceDetails/FilterViewController.swift
+++ b/Prox/Prox/PlaceDetails/FilterViewController.swift
@@ -23,6 +23,8 @@ class FilterViewController: UIViewController {
     private let topRatedSwitch = UISwitch()
     private let closeButton = UIButton()
 
+    private static var didShowPopularityToast = false
+
     private let filterLabels: [(PlaceFilter, String)] = [
         (.discover, Strings.filterView.discover),
         (.eatAndDrink, Strings.filterView.eatAndDrink),
@@ -159,6 +161,11 @@ class FilterViewController: UIViewController {
     }
 
     @objc private func didToggleRatings() {
+        if !FilterViewController.didShowPopularityToast {
+            FilterViewController.didShowPopularityToast = true
+            Toast(text: Strings.filterView.popularityToast).show()
+        }
+
         delegate?.filterViewController(self, didUpdateFilters: enabledFilters, topRatedOnly: topRatedSwitch.isOn)
     }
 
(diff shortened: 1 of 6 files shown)


Prox/Prox/PlaceDetails/FilterViewController.swift
```diff
@@ -23,6 +23,8 @@ class FilterViewController: UIViewController {
     private let topRatedSwitch = UISwitch()
     private let closeButton = UIButton()
 
+    private static var didShowPopularityToast = false
```
mcomella: Do we want to persist this to disk?

Prox/Prox/Widgets/Toast.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    func show() {
+        guard let window = UIApplication.shared.windows.first else { return }
+
+        let toast = UIView()
+        toast.alpha = 0
+        toast.backgroundColor = Colors.toastBackground
+        toast.layer.cornerRadius = 18
+        window.addSubview(toast)
```
mcomella: Is this guaranteed to draw on top?

Prox/Prox/Widgets/Toast.swift
```diff
@@ -0,0 +1,52 @@
...
+        label.text = text
+        label.textColor = Colors.toastText
+        label.font = Fonts.toast
+        label.numberOfLines = 0
+        toast.addSubview(label)
+
+        toast.snp.makeConstraints { make in
+            make.bottom.equalTo(window).offset(-30)
+            make.centerX.equalTo(window)
+            make.leading.greaterThanOrEqualTo(window)
```
mcomella: Why greaterThanOrEqual? It sounds like it could result in an ambiguous layout (I think greaterThan is typically used with optional constraints).

Same below for lessThan...

Prox/Prox/PlaceDetails/FilterViewController.swift
thebnich: Don't think so (at least not for Chicago).

Prox/Prox/Widgets/Toast.swift
thebnich: Yes (since it's on `window`).

Prox/Prox/Widgets/Toast.swift
thebnich: Because we also have the `label` constraint: `make.leading.trailing.equalTo(toast).inset(20)`

Combined, these constraints will make the toast grow with the label until it reaches the screen width, at which point they'll force a line break.

PR #625, opened 2017-03-09
Closes #623: Use latlong instead of address for routing

### LINKED ISSUE #623: Open directions in app rather than browser
Priority should be Google Maps (if installed), Apple Maps, Browser (but this should be the final fallback, since everyone should have Apple Maps installed).

liuche:
Looked at history and it seems like we've been using "address" for Apple Maps routing, but that's not necessary because we always have a lat/long. Switched to using latlong (instead of falling through to Google maps web if the address didn't exist).


liuche:
From the Apple Maps docs, the destination address needs to be "an address string that geolocation can understand" and lat/long falls into that category.

https://developer.apple.com/library/content/featuredarticles/iPhoneURLScheme_Reference/MapLinks/MapLinks.html

PR #630, opened 2017-03-09
Closes #544: Transition map view controller in from top

### LINKED ISSUE #544: Transition map view controller in from top

thebnich:
Copied from `FilterViewController`, using different views the for animations.

### DIFF
--- a/Prox/Prox/Map View/MapViewController.swift
+++ b/Prox/Prox/Map View/MapViewController.swift
@@ -4,6 +4,7 @@
 
 import UIKit
 import GoogleMaps
+import SnapKit
 
 private let fadeDuration: TimeInterval = 0.4
 
@@ -22,7 +23,7 @@ private let footerBottomOffset = Style.cardViewCornerRadius
 private let footerCardMargin = 16
 
 protocol MapViewControllerDelegate: class {
-    func mapViewController(didSelect: Place)
+    func mapViewController(_ mapViewController: MapViewController, didDismissWithSelectedPlace place: Place?)
 }
 
 enum MapState {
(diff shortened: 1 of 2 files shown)


Prox/Prox/Map View/MapViewController.swift
```diff
@@ -77,31 +81,50 @@ class MapViewController: UIViewController {
...
+        for subview in [container, mapView, searchButton, placeFooter, closeButton] as [UIView] {
             view.addSubview(subview)
         }
 
+        container.snp.makeConstraints { make in
+            make.leading.trailing.equalToSuperview()
+
+            showMapConstraints = [
+                make.top.equalToSuperview().constraint,
+                make.bottom.equalTo(bottomLayoutGuide.snp.top).constraint,
```
mcomella: Why bottom layout guide?

Prox/Prox/Map View/MapViewController.swift
```diff
@@ -131,22 +154,13 @@ class MapViewController: UIViewController {
     }
 
     @objc private func closeWithButton() {
-        resetMapToUserLocation(shouldAnimate: true)
-
-        // HACK: (no UX input) Delay the dismiss so the user can watch the animation and see what happens.
-        // TODO: we should change the delay depending on how far we animate.
-        // Note: I tried to add a listener for the end of the animation to dismiss but it complicated the code too much
-        // so here's this hack.
-        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + DispatchTimeInterval.milliseconds(500)) {
```
mcomella: As per irl, I don't think we should rm this.

thebnich: Talking to @antlam, he agrees that the return-to-position effect isn't worth the lag the user feels after pressing the close button, so keeping rm'ed.

Prox/Prox/Map View/MapViewController.swift
thebnich: Just copied from [what we were using before](https://github.com/mozilla-mobile/prox/pull/630/files#diff-8364cdee0b6d84d4da354f4b18cc16b5L116). I'll just change this to bottom.

PR #632, opened 2017-03-10
Closes #496: Use hostname to fetch app

### LINKED ISSUE #496: Sometimes clicking Yelp/TA opens in website instead of app (when app installed)
antlam has seen this and study participants have also commented on this.

iirc, kbrosnan may have found a phone at the all-hands that could reproduce.

liuche:
Instead of using the hard-coded website urls, extract the hostname and use that.


Prox/Prox/Utilities/OpenInHelper.swift
```diff
@@ -33,6 +34,15 @@ struct OpenInHelper {
         return true
     }
 
+    private static func extractHostname(fromHost host: String) -> String? {
+        let components = host.components(separatedBy: ".")
+        if components.count > 0 {
+            return components[1]
```
mcomella: So these are always guaranteed to be `www` URLs? Could there be multiple sub-domains?

Prox/Prox/Utilities/OpenInHelper.swift
```diff
@@ -33,6 +34,15 @@ struct OpenInHelper {
         return true
     }
 
+    private static func extractHostname(fromHost host: String) -> String? {
+        let components = host.components(separatedBy: ".")
+        if components.count > 0 {
```
thebnich: Nit: `components.count` will always have at least 1 item, so you could just change this to `return components(separatedBy: ".")[1]`

liuche: I think I'll keep it as is, though I'll document the assumption about the format of the url.

PR #633, opened 2017-03-10
Closes #558: Update ratings branding

### LINKED ISSUE #558: Use Branded UI for Ratings

mcomella:


Prox/Prox/Widgets/ProviderStarImage.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    /// Returns a width constraint that, when activated, constrains the image view to the asset size,
+    /// based on the given height.
+    ///
+    /// HACK: when we scale inside a UIImageView, the UIImageView keeps the size the of the original image
+    /// (rather than taking the scaled image size) so, in this case, the view is too wide and has whitespace
+    /// on either side of the image. For dev speed, we hardcode the asset sizes and set width constraints
+    /// with this method. A proper alternative would be to scale the image outside the image view
+    /// and set the image view with the scaled image.
+    func widthConstraint(for view: UIImageView, withViewHeight viewHeight: CGFloat) -> NSLayoutConstraint {
```
thebnich: This is pretty ugly. Consider importing [this](https://github.com/mozilla-mobile/firefox-ios/blob/716536a06c624dc9424e25edaca012dd50c49bbc/Shared/Extensions/UIImageExtensions.swift#L53) so we can keep this neatly contained in the image class.

Prox/Prox/Widgets/ProviderStarImage.swift
```diff
@@ -0,0 +1,82 @@
...
+        default:
+            log.warn("Unexpected Yelp score \(score). Returning nil.")
+            return nil
+        }
+
+        return img
+    }
+}
+
+struct TripAdvisorStarImageAccessor: ProviderStarImageAccessor {
```
thebnich: Why not make these subclasses of `UIImage`, with `init(forScore:)`?

Prox/Prox/Widgets/ProviderStarImage.swift
mcomella: lol, I don't know why I thought this would be simpler than scaling the image...

Prox/Prox/Widgets/ProviderStarImage.swift
mcomella: FYI: I opted to create static fn, like `ProviderStarImage.yelp(forScore` and pass around function references because I felt it was more flexible than tying these calls to a UIImage reference.

PR #635, opened 2017-03-10
Slide map view card when showing/hiding

thebnich:
Waiting on @antlam to confirm the new durations.


thebnich:
Going to skip this one for now since it changes the layout (antlam noticed the map is off-center) and the slide-in effect feels awkward with the contrasting slide-down map view animation.

PR #637, opened 2017-03-11
Event style for mini card in map view

thebnich:


Prox/Prox/Models/Place.swift
```diff
@@ -78,6 +79,7 @@ class Place: Hashable {
             self.wikipediaProvider = wikipediaProvider
             self.googleProvider = googleProvider
             self.customProvider = customProvider
+            self.isEvent = !(customProvider?.description ?? "").isEmpty
```
mcomella: nit: I'm not a fan of the readability of a bunch of syntax. I'd prefer if you used liuche's `String.isEmpty` method (the static version, not the instance one).

Prox/Prox/PlaceDetails/PlaceDetailsCardView.swift
```diff
@@ -327,7 +327,7 @@ class PlaceDetailsCardView: UIView {
 
         hoursView.iconView.isHidden = false
         hoursView.isPrimaryTextLabelHidden = false
-        hoursView.primaryTextLabel.text = primaryText
+        hoursView.primaryTextLabel.text = primaryText.capitalized
```
mcomella: I'm concerned this means we'll be capitalizing something we didn't mean to (e.g. error state)... But whatever.

Prox/Prox/Map View/MapViewCardFooter.swift
```diff
@@ -93,5 +135,21 @@ class MapViewCardFooter: ExpandingCardView {
...
+        if place.isEvent {
+            placeConstraints.forEach { $0.deactivate() }
+            eventConstraints.forEach { $0.activate() }
+        } else {
+            eventConstraints.forEach { $0.deactivate() }
+            placeConstraints.forEach { $0.activate() }
+        }
+        layoutIfNeeded()
+
+        guard let text = place.hours?.getEventTimeText(forToday: Date()) else { return }
```
mcomella: nit: slightly more readable if you destructure rather than using `text.0`

PR #638, opened 2017-03-11
No issue: assert Toast.show is called on main thread.

mcomella:
I experienced Exceptions otherwise.


mcomella:
Test code! I added a toast when we re-sorted places (so I could walk around and see when they're updated) which is called from the background thread.

Side note: I'm going to wait to land this until after we cut the final Chicago build so we don't crash on something that may not otherwise cause crashes.

thebnich:
Note that asserts are only enabled in debug builds, so that shouldn't be a problem.

Issue #584, opened 2017-03-02
"Check map for distance" instead of travel times on initial Discover category


mcomella:
I think this may be because we get travel times for *all* cards, which uses up our rates before we get to the Discover cards.

PR #643, opened 2017-03-16
No issue: PlacesProvider cleanup

mcomella:
Leftovers from #584.

### DIFF
--- a/Prox/Prox/Data/PlacesController.swift
+++ b/Prox/Prox/Data/PlacesController.swift
@@ -50,6 +50,7 @@ class PlacesProvider {
 
     init() {}
 
+    // only accessible for tests.
     convenience init(places: [Place]) {
         self.init()
         self.displayedPlaces = places
--- a/Prox/Prox/Data/PlacesController.swift
+++ b/Prox/Prox/Data/PlacesController.swift
@@ -72,12 +72,6 @@ class PlacesProvider {
         }
     }
 
-    func filterPlaces(enabledFilters: Set<PlaceFilter>, topRatedOnly: Bool) -> [Place] {
-        return placesLock.withReadLock {
-            return filterPlacesLocked(enabledFilters: enabledFilters, topRatedOnly: topRatedOnly)
-        }
-    }
-
     /// Callers must acquire a read lock before calling this method!
     /// TODO: Terrible name, terrible pattern. Fix this with #529.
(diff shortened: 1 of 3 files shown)


Prox/Prox/Data/PlacesController.swift
```diff
@@ -187,10 +180,20 @@ class PlacesProvider {
...
+    /// Returns a copy of the displayed places. Callers should not be locked.
     func getDisplayedPlacesCopy() -> [Place] {
+        return getCopy(ofPlaces: displayedPlaces)
+    }
+
+    private func getCopy(ofPlaces places: [Place]) -> [Place] {
         var placesCopy: [Place] = []
         placesLock.withReadLock {
-            placesCopy = Array(self.displayedPlaces)
+            placesCopy = Array(places)
```
thebnich: You don't need the `Array()` initializer since Swift arrays are already copy-on-write (I'd also remove `Copy` from the names since that's standard Swift behavior).

Prox/Prox/Data/PlacesController.swift
```diff
@@ -50,16 +50,6 @@ class PlacesProvider {
 
     init() {}
 
-    convenience init(places: [Place]) {
```
thebnich: This is still used in `PlacesProviderTests`.

Prox/Prox/Data/PlacesController.swift
mcomella: Wow, I had no idea... I wonder how many bugs I've introduced because I thought I was looking at the same reference.

mcomella: Thought: this means that since `allPlaces`/`displayedPlaces` are protected fields and Swift doesn’t have a concept of atomic assignment, every time we call a method with places as an argument, it has to be locked.

Otherwise, in theory, we could be in the middle of copying (for a fn call) when someone else modifies the Array.

(talking on Slack to figure out if this is actually an issue)

mcomella: Conclusion: structs (so Arrays) are copy-on-write, which the system handles the concurrency for, so there is no need to protect the resource, assuming it's always updated to a consistent state in a single operation.

mcomella:
@thebnich Please re-review. I changed some code to rely on the copy-on-write mechanics and I want to make sure it's both clear and correct.

PR #654, opened 2017-03-29
Show local events

liuche:


Prox/Prox/Models/Event.swift
```diff
@@ -4,11 +4,10 @@
 
 import Foundation
 
-private let idPrefix = "proxevent-"
-
 /// An event. Properties are modelled after an Eventbrite Event:
 ///   https://www.eventbrite.com/developer/v3/response_formats/event/#ebapi-std:format-event
 /// This is likely to change when we add events from other providers.
+
```
mcomella: nit: not worth undoing but no ws.

